PR #16, opened 2024-06-18
Reorganize to prep for applying scientific-python cookie

### LINKED ISSUE #5: add CI/CD workflows for publishing, linting, testing, and documentation generation
We should adopt in this repo the best practices for scientific python programming. A streamlined way to do this is to apply this cookiecutter: https://github.com/scientific-python/cookie. This would add:

- GitHub workflows for CI and CD. Each time a GitHub Release is created, it triggers the build, test, and publishing of wheels.
- Extraction of package version from source control by leveraging setuptools-scm. The generated source distribution would include the relevant version metadata.
- Pre-commit hooks for checking style. To start with, I plan to add linting exceptions to prevent sweeping style changes. I can make only the smaller style changes.
- Dependabot configuration. This automates a lot of routine maintenance

The major benefit of this that openlifu can be more easily maintained and automatically published to PyPI upon release. This brings it further into the larger python ecosystem by allowing other packages to mark it as a dependency.

ebrahimebrahim:
Re-organize the repo in order to prepare for the PR #17 that will resolve #5 by applying the scientific-python cookiecutter to add CI.

- Move project files into a `src/` subdirectory
  - Remove the top level __init__ from the repo, which was wrongly making the top level appear to be a python package.
  - Move `pyfus` to `src/pyfus`. This way when you are working in the repo  directory python will not try import things from the current directory.  The should be no confusion about where imports are coming from.
  - Remove [tool.setuptools.packages.find] from the pyproject.toml. It is not needed -- I believe setuptools already finds pacakages by looking for  the` __init__.py `s
- Rename import package `pyfus` --> `openlifu`
- Apply automatic formatting fixes: Files updated using `end-of-file-fixer`, `mixed-line-ending`, `trailing-whitespace`,
`blacken-docs`, and `prettier` pre-commit hooks. These are the easy non-sweeping style fixes that will make the linting pass once we introduce it. For the other style fixes, they are going to be configured to be ignored by the linter and we can add those rules back in over time in separate PRs.
- Add empty file `docs/_static/.gitkeep`: Sphinx is configured to look for this directory, this change helps suppress a sphinx warning
- Rename test_notebooks->notebooks. Unit tests will be brought in soon and we don't want the test_notebooks
to look like they are part of that.


src/openlifu/seg/seg_methods/tissue.py
peterhollender, line 1: @ebrahimebrahim do you know why this change and `water.py` are showing up as whole new files instead of renames+ modification like the others?

peterhollender (approved):
Looks good. I see that it's just moving/renaming/adjusting line endings.

src/openlifu/seg/seg_methods/tissue.py
ebrahimebrahim: Hmm I'm not certain but it might be due to the length of the file. The file is short, so because 9f23b5b3a9c720a88c551e6cc1db74078ac10ba2 had to both rename (move) the file and edit it (changing pyfus to openlifu) it was treated by git as a large proportion of the file changing. When git sees a rename and a large proportion of changes I think it treats that as a delete-and-add rather than as a move

ebrahimebrahim:
(Disabling status checks to merge this before bringing in #17 which actually introduces status checks. Will re-enable status checks after merging.)

Issue #7, opened 2024-06-17
k-wave-python specific branch needed


ebrahimebrahim:
A [custom branch of k-wave-python](https://github.com/peterhollender/k-wave-python/tree/fix_cached_windows_binaries) is currently needed. (EDIT: I think it's actually the `master` branch of https://github.com/peterhollender/k-wave-python that we want, but the point is that it's this fork and not the k-wave-python that is published to the package index)

This currently requires us to put a direct reference in the `dependencies` section of the  `pyproject.toml`.

[Direct references ](https://packaging.python.org/en/latest/specifications/version-specifiers/#direct-references) are better avoided, and we will have to find some way around this before we will be able to publish to PyPI.

peterhollender:
Ok, the `master` branch of the original https://github.com/waltsims/k-wave-python now has been patched to fix the import bug, and we do not need to use the custom fork. As of 2024-06-24, the version posted to PyPI has not yet reflected the merge.

ebrahimebrahim:
That's great! 

For now we can switch to their master branch but I'll leave this issue open until they do another release -- at that point we can remove the direct reference and unblock our own publlishing to pypu

ebrahimebrahim:
(I've subscribed to be notified by github when they do a release so I can finish this issue as soon as that's done)

PR #20, opened 2024-06-25
Switch to master branch of k-wave-python

ebrahimebrahim:
This is part of #7 but we still need them to release the new version before it's solved


peterhollender (approved):
Tested and works. Will keep an eye out for the PyPI version of `k-wave-python` to get updated so that we can move the dependency away from the GitHub hosted version.

Issue #15, opened 2024-06-17
make notebooks play more nicely with version control


ebrahimebrahim:
I'd like to suggest using the wonderful tool [Jupytext](https://jupytext.readthedocs.io/) for the ipynb notebooks. The tool allows you to save a notebook as an ordinary python `.py` text file, with no cell output and with the cell metadata encoded inside code comments. This text file is easily human readable and works well with git version control. Meanwhile, jupytext lets you keep a local `.ipynb` notebook, and seemlessly pair it with the

If this suggestion is well received then the thing to do is:
- pair the ipynb files with some jupytext `.py` plain text format
- add a `README.md` inside the notebooks directory explaining how to use these plain text notebooks and open them as notebooks
- replace the `ipynb` files by `py` files in the git repo

No more reviewing PRs with unreadable notebook metadata changes!
No more accidentally committing cell output with a giant image!
No more irrelevant changes to the working tree every time a notebook is re-run and the cell IDs change!

PR #21, opened 2024-07-01
P/fix kwave if

### LINKED ISSUE #9: test_first notebook throws "unexpected keyword argument 'grid_weights'"
The [test_first notebook](https://github.com/OpenwaterHealth/OpenLIFU-python/blob/46d73e8d70886911890a8e20351ea15d27ed8460/test_notebooks/test_first.ipynb) contains this error for me:

```txt
TypeError: kWaveArray.get_distributed_source_signal() got an unexpected keyword argument 'grid_weights'
```

in reference to this line:
https://github.com/OpenwaterHealth/OpenLIFU-python/blob/46d73e8d70886911890a8e20351ea15d27ed8460/pyfus/sim/kwave_if.py#L64

### LINKED ISSUE #22: Run simulation always tries to run GPU code
The CPU version of the code is not called because `kwave_if\run_simulation` creates a `SimulationExecutionOptions` object with `is_gpu_simulation=True` hardcoded.

peterhollender:
Closes #9 by removing the gridweights interface altogether. I don't think that the implementation on my fork is going to be how a first-party fix gets completed, so it's not really worth keeping, and the performance on `master` seems to have improved at least somewhat, perhaps from another commit.

Also incidentally fixes an issue where non-GPU based simulation wasn't getting called, which closes #22


notebooks/test_first.ipynb
```diff
@@ -138,9 +746,37 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 12,
    "metadata": {},
-   "outputs": [],
+   "outputs": [
```
ebrahimebrahim, line 751: Did you mean to include cell outputs in this commit? It will be better clear all cell output to avoid bloating the repo.

(This will stop being an issue once we do #15 :smile: -- I can prioritize that next if interested)

ebrahimebrahim (changes requested):
Nice, thanks!

Commits are separated out nicely with each commit doing one thing, that's great

To start following our new guidelines in [CONTRIBUTING.md](https://github.com/OpenwaterHealth/OpenLIFU-python/blob/main/.github/CONTRIBUTING.md#commit-and-pull-request-expectations):
> - Every commit should reference the issue number with which it is associated.
> - Commits should be reasonably granular and semantically atomic.
> - Pull requests should not be squashed upon merging.

can you make sure that issue numbers are referenced in the title and body of each commit? It looks like two of the three commits are associated with #9. For the commit about adding the ability to disable gpu, you would have to open a new issue with a description of the thing being done (super short description is fine -- we just want to track which changes are for which requirement) and then reference that new issue in the commit.

For editing commit messages I usually do this by an interactive rebase -- let me know on discord or wherever if you could use help with that

notebooks/test_first.ipynb
peterhollender: I did an interactive rebase and chose to reword the commits, but now I'm seeing a second set of them tracked in this PR... guessing I didn't fix that correctly

ebrahimebrahim: Interactive rebase will change the commit hashes, so that's correct actually!
You would have to force-push the new commits to replace the previous commit history:
```
git push --force-with-lease
```

Regarding removing cell outputs: adding an additional commit that removes the cell output still leaves the output in the commit history, which continues to bloat the repo. What's needed ultimately is to change that commit that introduces the output and remove the output from it. Alternatively, the two commits (the one that introduces cell output and the one that removes it) can be squashed to achieve that same effect. Happy to hop onto a call if I can help in any way!

ebrahimebrahim: (It looks like you did force-push so that's good.)

ebrahimebrahim (approved):
It looks good! Thanks for your patience with the new software processes :)

I am adding "also closes #22" to the PR description. Feel free to merge

PR #78, opened 2024-07-09
Make k-wave-python a direct dependency

### LINKED ISSUE #7: k-wave-python specific branch needed

ebrahimebrahim:
Closes #7

It was previously indirect becasue of a windows issue in the main codebase that had yet to be released.

The new release is said to support MacOS so we open that up for testing in the CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,7 +43,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: ["3.9", "3.12"]
-        runs-on: [ubuntu-latest, windows-latest]
+        runs-on: [ubuntu-latest, windows-latest, macos-latest]
 
     steps:
       - uses: actions/checkout@v4
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,12 +42,8 @@ dependencies = [
   "nibabel",
   "sphinx",
   "ipykernel",
-  "k-wave-python @ git+https://github.com/waltsims/k-wave-python"
+  "k-wave-python>=0.3.4",
 ]
-
-# We should remove this section once we resolve the k-wave-python issue
-[tool.hatch.metadata]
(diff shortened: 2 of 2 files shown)


peterhollender (commented):
Could we configure the project so that it runs an `import kwave` as part of the installation process, which will trigger the download of the binaries if they aren't found on the system? That way the dependency gets fully installed.

something like:
 `python -c "import logging; logging.basicConfig(level=logging.INFO); import kwave"`

ebrahimebrahim:
> Could we configure the project so that it runs an `import kwave` as part of the installation process, which will trigger the download of the binaries if they aren't found on the system? That way the dependency gets fully installed.
> 
> something like: `python -c "import logging; logging.basicConfig(level=logging.INFO); import kwave"`

Great idea. As I look into it, it seems not as straightforward as I thought. Let me keep digging for a bit

ebrahimebrahim:
Will continue looking into this, but if we decide that it should be an issue on k-wave-python then we can also open that issue and merge this PR. Let me know

peterhollender:
I think we open the issue on k-wave-python for more verbose communication during first time download, and solve this during custom app deployment for ours. We can resolve this issue.

peterhollender (approved):

PR #86, opened 2024-08-21
Support serialization for Protocols and Transducers (#85)

### LINKED ISSUE #85: Create methods for serializing/deserializing Protocols and Transducers
This is needed to support the parameter node wrapping for https://github.com/OpenwaterHealth/SlicerOpenLIFU/issues/57, allowing protocols and transducer to persist in slicer scenes.

For making a parameter node serializer it should also be possible to create a default  Protocol and a default Transducer

ebrahimebrahim:
Close #85 

- Make `Protocol.from_dict` work; it was previously broken due to the `sim_setup` step
- Add `from_json` and `to_json` methods to `Protocol`, and a unit test that ensures they go back and forth consistently
- Getting this to pass required some fixes to `SegmentationMethod` dict conversion:
  - Slightly narrowed the set of fields grabbed by `SegmentationMethod.to_dict`
  - Ensured that `SegmentationMethod.from_dict` knows how to construct `Material` objects when it needs to
- Add `from_json` and `to_json` methods to `Transducer`, and a unit test that ensures they go back and forth consistently
- Make sure it is possible to create a default protocol; it was not possible because it was trying to internally instantiate an abstract focal pattern

I see there is a custom `JSONEncoder` in `openlifu.util.json`, and I considered adding functionality there or re-using that in some way. But I opted not to because there isn't an openlifu-wide consistent method of serializing and deserializing to files; we are currently a bit all over the place with some classes doing their own thing and other classes using `openlifu.util.json`. So for now I'd rather just have `Protocol` and `Transducer` define their own personal serialization functions so that we can use them in SlicerOpenLIFU, and then later if there is interest we can revamp the serialization functionality across all of openlifu to be a bit more streamlined and consistent.

### DIFF
--- a/src/openlifu/bf/__init__.py
+++ b/src/openlifu/bf/__init__.py
@@ -1,5 +1,5 @@
 from .delay_methods import DelayMethod
 from .apod_methods import ApodizationMethod
-from .focal_patterns import FocalPattern
+from .focal_patterns import FocalPattern, SinglePoint, Wheel
 from .pulse import Pulse
 from .sequence import Sequence
--- a/src/openlifu/plan/protocol.py
+++ b/src/openlifu/plan/protocol.py
@@ -11,7 +11,7 @@ class Protocol:
     description: str = ""
     pulse: bf.Pulse = field(default_factory=bf.Pulse)
     sequence: bf.Sequence = field(default_factory=bf.Sequence)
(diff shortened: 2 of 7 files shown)


peterhollender (approved):
Looks good. Agreed that we will want to clean up and streamline serialization in the future.

PR #99, opened 2024-09-08
Add to/from json for Points

### LINKED ISSUE #98: Enable serialization of points
In SlicerOpenLIFU it will be good to save the focal pattern points that are generated during treatment planning into an object that is tracked by a parameter node. It would be convenient to represent them directly as `openlifu.Point`s rather than transforming them to some other representation, and this can be most easily achieved if `Point`s can be reliably serialized into and deserialized from strings. This is needed by https://github.com/OpenwaterHealth/SlicerOpenLIFU/issues/61

ebrahimebrahim:
Close #98

### DIFF
--- a/src/openlifu/geo.py
+++ b/src/openlifu/geo.py
@@ -4,6 +4,7 @@ from dataclasses import dataclass, field
 from openlifu.util.units import getunitconversion
 import copy
 import vtk
+import json
 
 @dataclass
 class Point:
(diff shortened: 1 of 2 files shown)


peterhollender (commented):
I think we had talked about removing the `color` and `radius` fields? I guess that would be a separate change from just the json compatibility?

ebrahimebrahim:
We could include it in this PR, do you want to? We don't have that many unit tests yet so it may make sense to put off changes like that till the unit tests are more fleshed out. IDK how breaking that change would be

peterhollender:
> We could include it in this PR, do you want to? We don't have that many unit tests yet so it may make sense to put off changes like that till the unit tests are more fleshed out. IDK how breaking that change would be

Ok, that makes sense. It isn't actively hurting anything to have those fields, and it has defaults for them, so we can handle that later.

peterhollender (approved):
Works for me. We can modify the attributes later.

PR #101, opened 2024-09-11
Fix subject creation (#90)

### LINKED ISSUE #90: Fix subject creation
Database.add_subject doesn’t work because Subject.to_file fails to create directories when it needs to.

ebrahimebrahim:
Close #90


peterhollender (approved):
looks good

PR #109, opened 2024-09-16
Update session loading to use transducer_id and array_transform 

sadhana-r:


ebrahimebrahim:
Thank you for starting this and identifying what needed to be done to load the new data
Closing this PR in favor of #110

PR #112, opened 2024-09-17
Fix vtk warning when loading a transducer model

### LINKED ISSUE #108: There is a vtk warning when loading a transducer model
After loading a transducer model there is a warning

```
[VTK] Bad table range: [0, -1]
```

I haven't noticed any problems caused by this yet.

The fix for this could be in openlifu, we'll see. This is what we use to get the model: https://github.com/OpenwaterHealth/OpenLIFU-python/blob/06f8756b254d627f6f6919d5ed7b705d15a99a21/src/openlifu/xdc/transducer.py#L73

medmssaouri:
Close #108

### DIFF
--- a/src/openlifu/xdc/transducer.py
+++ b/src/openlifu/xdc/transducer.py
@@ -47,10 +47,9 @@ class Transducer:
     def draw(self,
              units=None,
              transform=True,
-             facecolor=[0,1,1],
-             facealpha=0.5):
+             facecolor=[0,1,1,0.5]):
         units = self.units if units is None else units
-        actor = self.get_actor(units=units, transform=transform, facecolor=facecolor, facealpha=facealpha)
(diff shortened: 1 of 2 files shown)


ebrahimebrahim:
Nice, thank you! A couple of things:
- Instead of _removing_ the option to color, can we set it so that when the color arguments are passed as `None` then they are the color and/or alpha are not set on the polydata. This way coloring the polydata remains an option, but we can turn it off in SlicerOpenLIFU.
- Can you add a unit test that checks that a polydata can be constructed? It could be added to `test_transducer`. Even better if you can test that the color options functionality works after your change

medmssaouri:
@ebrahimebrahim  Could please take a second look, thanks

ebrahimebrahim (approved):
I gave it a look and tested it before the latest force-push, but assuming it still works after that I think it's good! Feel free to merge if you feel it's ready.

PR #116, opened 2024-09-19
Implement solution class with serialization capabilities

### LINKED ISSUE #103: Implement Solution class
Add Solutions to openlifu-python, referencing the matlab code for what needs to go into them, and including the needed serialization functionality.

This is just putting together the class mostly as a data structure with serialization -- all the analysis functionality we will worry about later in #104 and #105.

[Here is the definition of Solution in the matlab code.](https://github.com/OpenwaterHealth/opw_neuromod_sw/blob/main/%2Bfus/%2Btreatment/Solution.m)

Additional note from Peter:
> `fus.treatment.Plan.m` (equivalent to `openlifu.Protocol`) is also a useful reference, as it also has some of the planning responsibilities that go into programmatically generating the Solution

### LINKED ISSUE #119: Create base class for common dict conversion code
In reference to the basic implementations of `to_dict` and `from_dict methods`:

> This function member seems to repeat itself in many other dataclasses. Better to create a mother class that implements them (`to_dict` and `from_dict`).

_Originally posted by @ltetrel in https://github.com/OpenwaterHealth/OpenLIFU-python/pull/116#discussion_r1770943414_

### LINKED ISSUE #120: Use enum classes for "on conflict" options
In reference to the various "on conflict" options:

> Use enum classes here for type safety instead of raw strings. If the goal is to only port the matlab code then please ignore this (since this is what was written in matlab).

_Originally posted by @ltetrel in https://github.com/OpenwaterHealth/OpenLIFU-python/pull/116#discussion_r1770952869_

ebrahimebrahim:
Close #103
Close #119 
Close #120 

Implements the solution class, enabling json and file serialization. Simulation outputs (`xarray.Dataset`) are serialized using [the netCDF format, recommended by xarray](https://docs.xarray.dev/en/stable/user-guide/io.html#netcdf). The needed database functionality is also added.

## remaining tasks

- [x] rebase to main after #110 is merged


ebrahimebrahim:
If trying this with example databases then it should work with db-simple-example-v04 and db-extended-example-v03 from our [shared data store](https://drive.google.com/drive/folders/1bldoP9VxE9SHyAB0bVAJ2l6WUuTxz3Yu).

src/openlifu/bf/sequence.py
```diff
@@ -38,18 +40,12 @@ def from_dict(d):
         :param d: Dictionary of the sequence parameters
         :returns: Sequence object
         """
-        return Sequence(pulse_interval=d["pulse_interval"],
-                        pulse_count=d["pulse_count"],
-                        pulse_train_interval=d["pulse_train_interval"],
-                        pulse_train_count=d["pulse_train_count"])
+        return Sequence(**d)
 
     def to_dict(self):
```
ltetrel, line 45: This function member seems to repeat itself in many other dataclasses. Better to create a mother class that implements them (`to_dict` and `from_dict`).

src/openlifu/db/database.py
```diff
@@ -206,6 +205,29 @@ def write_volume(self, subject, volume, on_conflict="error"):
...
+        if solution.id in solution_ids:
+            if on_conflict == "error":
+                raise ValueError(f"Solution with ID {solution.id} already exists in the database.")
+            elif on_conflict == "overwrite":
+                self.logger.info(f"Overwriting solution with ID {solution.id} in the database.")
+            elif on_conflict == "skip":
+                self.logger.info(f"Skipping solution with ID {solution.id} as it already exists.")
+                return
+            else:
+                raise ValueError("Invalid 'on_conflict' option. Use 'error', 'overwrite', or 'skip'.")
```
ltetrel, line 220: Use enum classes here for type safety instead of raw strings. If the goal is to only port the matlab code then please ignore this (since this is what was written in matlab).

src/openlifu/db/database.py
```diff
@@ -456,26 +469,18 @@ def load_session_info(self, subject_id, session_id):
             raise FileNotFoundError(f"Session file not found for ID: {session_id}")
 
 
-    def load_solution(self, session, protocol_id, target_id):
-        solution_filename_json = self.get_solution_filename(session.subject_id, session.id, protocol_id, target_id, ext="json")
-        solution_filename_mat = self.get_solution_filename(session.subject_id, session.id, protocol_id, target_id, ext="mat")
+    def load_solution(self, session:Session, solution_id:str) -> Solution:
```
ltetrel, line 472: In the matlab code, the function is defined as:
```matlab
[solution, output] = load_solution(self, session, plan_id, target_id) 
```
Currently It is missing the `output`

tests/resources/example_db/subjects/example_subject/sessions/example_session/solutions/example_solution/example_solution.nc
ltetrel, line 1: Should we store the test data in a remote storage instead of github ? I am not sure since those are just a bunch of json files. If at one point we need to store medical data, we will have no choice.

tests/test_solution.py
```diff
@@ -0,0 +1,69 @@
...
+import numpy as np
+import pytest
+import xarray
+from helpers import dataclasses_are_equal
+
+from openlifu import Point, Pulse, Sequence, Solution
+
+
+@pytest.fixture()
+def example_solution() -> Solution:
```
ltetrel, line 13: Why not have this into a file, like you did for [tests/test_database.py](tests/test_database.py) ?

src/openlifu/plan/solution.py
```diff
@@ -1,7 +1,169 @@
...
+    # than having their own unique solution ID. We do have unique solution IDs so it's possible we don't need
+    # this target attribute at all here. Keeping it here for now just in case.
+    target: Optional[Point] = None
+    """The ultimate target of this sonication. This sonication solution is focused on one focal point
+    in a pattern that is centered on this target."""
+
+    # In the matlab code the simulation result was saved as a separate .mat file.
+    # Here we include it as an xarray dataset.
+    simulation_result: xarray.Dataset = field(default_factory=xarray.Dataset)
+    """The xarray Dataset of simulation results"""
```
ltetrel, line 68: Are you talking about what was previously `solutions/example_plan/example_target.mat` ?

ltetrel, line 68: Knowing that the developments targets a slicer app, should we not use a vtk data model here ? With the vtk python interface we should be able to define the different arrays into vtk field-data.
Or we want this lib to be more pythonic (with xarrays) ?

ltetrel (commented):
All the tests passed without issues.

tests/resources/example_db/subjects/example_subject/sessions/example_session/solutions/example_solution/example_solution.nc
ebrahimebrahim: This is a very small data file just for testing so I think it's okay here. For larger data storage and versioning we are still working on a solution. Sadhana was looking into setting up DVC, so that might be the way we will do it, but the unit test data we can keep super small and just commit it.

tests/test_solution.py
ebrahimebrahim: The goal of these tests is to test the functionality of the Solution class features independently, not necessarily file i/o. On the other hand `test_database` is focused on the file i/o aspect.

src/openlifu/plan/solution.py
ebrahimebrahim: I believe s. The `simulation_result` would contain the `ds` object [returned by `run_simulation`](https://github.com/OpenwaterHealth/OpenLIFU-python/blob/68166a29cbe564630b7c9933f36692060a1bb509/src/openlifu/sim/kwave_if.py#L127)

So this would be an xarray Dataset containing "pnp" (peak negative pressure), "ppp" (peak positive pressure), and "ita" (time average intensity) arrays defined over the simulation grid.

src/openlifu/db/database.py
ebrahimebrahim: Maybe you can check me on this but I think the `output` is the the `Solution.simulation_output` xarray here. In the  python library it is an attribute of `Solution` rather than a separate object. See [my answer below](https://github.com/OpenwaterHealth/OpenLIFU-python/pull/116#discussion_r1771330189) about what is in `Solution.simulation_output`

src/openlifu/plan/solution.py
ebrahimebrahim: Indeed we want this lib to be more pythonic :smile:

ebrahimebrahim:
@ltetrel appreciate the review! please have another look

src/openlifu/db/database.py
ltetrel: Indeed I just checked, it will also contain the `analysis` output.

ltetrel (approved):
LGTM,

Issue #91, opened 2024-09-04
Enable adding a volume to a subject


ebrahimebrahim:
Appending a volume ID to the list of volume IDs in a subject json is easy, and could be done by just reading the subject, modifying it, and then writing it back using Subject.to_file (once #90 is done). However this is not the right level of api to expose. SlicerOpenLIFU for example should not normally be relying on the file I/O that is internal to openlifu.Database. So we really need to use `Databse.add_volume`. But the latter is very broken: it assumes there is a type (perhaps called “Volume”) that has attributes “id” and “to_json”. The “to_json” was expected to give the actual volume data, which could then be written out to a json file. Several problems: there is no such “Volume” type, there is no `Database.write_volume_ids`, and of course json is not a good way to represent raw volume data on disk. 

`Database.add_volume` should be fixed, but this is a nontrivial fix that touches on openlifu.Database as a whole.

Depends on #90

ebrahimebrahim:
To clarify this a bit, the ultimate goal is to have a method `Database.add_volume` with inputs

- subject to add the volume to
- path to a volume file (could a .nii file or anything really -- openlifu is not responsible for volume file i/o so it doesn't need to worry what's the format)
- volume id and any other metadata that gets saved to that volume's json file. I think we should save volume's ID, name, and filename only as metadata. Not "dims" because any decent medical volume data format would encode that internally

Then `Database.add_volume` should

- add the volume id to the subject and write the subject
- copy the volume file into the subject's volumes folder (keeping its filename the same -- the metadata json will point to that filename)
- write the volume metadata to a json file

There should also be a simple `Database.get_volume_info(self, subject, volume_id)` method that returns the dictionary of metadata along with the _absolute_ filepath to the volume file.

ebrahimebrahim:
Closed via #123

PR #123, opened 2024-09-25
Issue 91

medmssaouri:


medmssaouri:
@ebrahimebrahim Could you take a look at this PR.
Note: i removed load_volume_attrs because it is not used in the project, please let me know whether you have in mind a case where it could be needed that that i keep it
Thanks

src/openlifu/db/database.py
```diff
@@ -377,22 +392,11 @@ def load_subject(self, subject_id, options=None):
         self.logger.info(f"Loaded subject {subject_id}")
         return subject
 
-    def load_volume(self, subject, volume_id):
-        raise NotImplementedError("Volume is not yet implemented")
-        volume_filename = self.get_volume_filename(subject.id, volume_id)
+    def load_volume(self, subject_id, volume_id):
```
ebrahimebrahim, line 395: This should not be called `load_volume` because that sounds like it is  loading a volume file, but it is not. It is loading just metadata and a filepath. Maybe `load_volume_info` or `get_volume_info`

src/openlifu/db/database.py
```diff
@@ -377,22 +392,11 @@ def load_subject(self, subject_id, options=None):
...
+        volume_filename = self.get_volume_filepath(subject_id, volume_id)
         volume = Volume.from_file(volume_filename)
         return volume
```
ebrahimebrahim, line 398: It's not enough to just return the volume metadata, we need to also get the absolute file path of the volume file (see https://github.com/OpenwaterHealth/OpenLIFU-python/issues/91#issuecomment-2365232909 ). The user needs to know where the volume file lives in order to load it themselves.

The way to get the path to the volume file is to use the path to the volume metadata json, take the parent directory of that, and then use the `filename` that should also be stored in the volume info json

src/openlifu/db/database.py
```diff
@@ -377,22 +392,11 @@ def load_subject(self, subject_id, options=None):
...
 
-    def load_volume_attrs(self, subject, volume_ids=None):
```
ebrahimebrahim, line 386: Yes it's okay to delete this, thanks!

src/openlifu/db/volume.py
```diff
@@ -0,0 +1,73 @@
+import json
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Dict, Optional
+
+from openlifu.util.json import PYFUSEncoder
+
+
+@dataclass
+class Volume:
```
ebrahimebrahim, line 10: I wasn't thinking that a new class would be needed for this, but I see that it is pretty nice to have a class. Still, I think it should not be called Volume because it actually just references a volume file rather than being an actual volume. Maybe `VolumeInfo`?

If we do add this class then I think in this PR we should unit test all the functionality. So `to_json`, etc. should all be tested. The suit of tests could be in a `test_volume.py` file

But we could also get by without having this class at all. All that `write_volume` needs to do is write a json file from a dict with `id`, `name`, and `filename`, and this can be done directly in the `Database` class. And then `Database.load_volume_info` (or `Database.get_volume_info`) would just have to read the json file and return that as a dict, as well as returning the filepath to the volume file.

src/openlifu/db/volume.py
```diff
@@ -0,0 +1,73 @@
...
+    """
+    Class representing a volume
+
+    ivar id: ID of the volume
+    ivar name: Name of the volume
+    """
+    id: Optional[str] = None
+    name: Optional[str] = None
```
ebrahimebrahim, line 18: id and name are good but there needs to be another attribute: `filename`. This is needed by Database in order to later recover the volume path

tests/resources/test_db_files/example_volume_2.nii
ebrahimebrahim, line 1: I think you can do the test without introducing this file. I will explain in the test

tests/test_database.py
```diff
@@ -135,6 +135,24 @@ def test_get_solution_ids(example_database:Database, caplog):
...
+    subject = "example_subject"
+    volume_id = "example_volume_2"
+    volume_name = "EXAMPLE_VOLUME_2"
+    volume_data_path = Path(__file__).parent/"resources/test_db_files/example_volume_2.nii"
+    example_database.write_volume(subject, volume_id, volume_name, volume_data_path)
+    assert(example_database.get_volume_ids("example_subject") == ["example_volume", "example_volume_2"])
+
+    volume_filepath = example_database.get_volume_filepath("example_subject", "example_volume_2")
+    assert(volume_filepath.name == "example_volume_2.json")
+    assert((volume_filepath.parent/"example_volume_2.nii").exists())
```
ebrahimebrahim, line 155: I think you can do this test without introducing the `example_volume_2.nii` resource. The test doesn't need a real nii file, it can just be an empty file. Add a parameter `tmp_path:Path` to `test_write_volume`, similar to how it's done in the `example_database` fixture. This `tmp_path` parameter is a special name for pytest, it will be a temporary directory that you can use just for this test. You can use [`Path.touch`](https://docs.python.org/3/library/pathlib.html#pathlib.Path.touch) to just make an empty file that is then passed as the volume file, and you can give it an nii extension or whatever.

src/openlifu/db/database.py
```diff
@@ -300,6 +303,18 @@ def get_session_ids(self, subject_id):
             self.logger.warning("Sessions file not found for subject %s.", subject_id)
             return []
 
+    def get_volume_ids(self, subject_id):
```
ebrahimebrahim, line 306: Better to unit test this since we are adding it here in this pr. Similarly for `write_volume_ids`

tests/resources/example_db/subjects/example_subject/volumes/example_volume/example_volume.json
ebrahimebrahim, line 1: This still contains `dims` and `dim_names` which are no longer part of the volume metadata format

ebrahimebrahim (changes requested):
Thanks for the work on this!

- everything added should be unit tested, or at least almost everything
- the critical thing to include will be some way of getting the absolute filepath to the volume file, so that we know what to tell Slicer to load later. remember that the volume filename need not match the name or id of the volume, so filename will have to be stored somewhere
- it may be simpler to do this without introducing a new class just to package together id+name+filename, but it's up to you how to approach it. a class can be nice ...I am not sure about it

medmssaouri:
@ebrahimebrahim Please give this another look

ebrahimebrahim (approved):
Excellent work and thank you for being flexible and addressing all the review comments!

One small thing is I just noticed that volume IDs were being listed in the subject file: https://github.com/OpenwaterHealth/OpenLIFU-python/blob/ae84cbb9f410ad3fb874e4e4feac58781d93044d/tests/resources/example_db/subjects/example_subject/example_subject.json#L4

I like your way of doing it in a volumes.json better, just because it lines up with the way the rest of the database system works.

I will just remove that line and then merge the PR myself. Thanks again!

PR #130, opened 2024-10-08
Add approval information to Sessions and Solutions

### LINKED ISSUE #95: Add approval information to Sessions
Add to `Session` definitions the approval information that should be stored there. I think it is just whether the virtual fitting result was approved. Plans can be "approved" in SlicerOpenLIFU but that just gets saved to the SlicerOpenLIFUPlan object, which doesn’t ever go the database.

[Relevant slide](https://docs.google.com/presentation/d/1yH7Nh0XqOb4q_BCZdTR4L2R6TGPShxaMDx2XHCmvxSg/edit#slide=id.g2f2d4b7456b_0_290) from SlicerOpenLIFU treatment workflow

Besides code changes, the example data should also be updated.

Depends on #92 

---

More detail:

- Add an `Optional[str]` to `openlifu.Session` called `virtual_fit_approval_for_target_id`. It's `None` by default, which means no virtual fit is approved. And otherwise the value it takes on is the ID of the approved target.
- When you save a session, if the target that was approved isn't in the list, an error should be raised.
- Add a `bool` to `openlifu.Session` called  `transducer_tracking_approved`, `False` by default.
- Add a `bool` to `openlifu.Solution` called  `approved`, `False` by default.

ebrahimebrahim:
Close #95 

The new attributes to pay attention to are

- `Session.virtual_fit_approval_for_target_id` (to be used in https://github.com/OpenwaterHealth/SlicerOpenLIFU/issues/84 -- see **Virtual fit approval**)
- `Session.transducer_tracking_approved` (to be used in https://github.com/OpenwaterHealth/SlicerOpenLIFU/issues/113 -- see **Transducer tracking approval**)
- `Solution.approved` (to be used in https://github.com/OpenwaterHealth/SlicerOpenLIFU/issues/113 -- see **Solution approval**)

@peterhollender I am having you review this one so that you get a chance to look at the above links and see how approvals are proposed to work. TY!


src/openlifu/db/database.py
peterhollender, line 1: I'm wondering if the virtual fit approval should be a list of target ids, or maybe it shouldn't actually live in the session itself? I'm trying to think through what happens if you have multiple targets. I'd think that would generate multiple solutions, and that each of those solutions could have the sequence of approvals stored for _that solution's target_. Like the solution would include a "virtual fit array transform" and a "measured array transform" or something. I'm also ok with having sessions be initially restricted to a single target, which could streamline some things.

tests/resources/example_db/subjects/example_subject/sessions/example_session/example_session.json
peterhollender, line 1: Yeah, as I look at this, I think we should constrain sessions to single targets for now, or at least have an "active" target. And then implement the business logic that clears the approvals if you change the active target.

peterhollender (approved):

Issue #96, opened 2024-09-04
Define `Run`s


ebrahimebrahim:
Add to openlifu the concept of a `Run` and the ability to serialize them to and from json. `Run`s should be things that are saved under a particular session, so also add a method to the Database class to save a run for a particular session, and update the `Session` definition. Like transducers (and [protocols](https://github.com/OpenwaterHealth/OpenLIFU-python/issues/93)), sessions should save the ID referencing the run and not the run itself.

The `Database` should also be able to retrieve runs by ID.

See also [SlicerOpenLIFURun](https://github.com/OpenwaterHealth/SlicerOpenLIFU/issues/85) for more context as to why the `Run` concept is desired.

ebrahimebrahim:
Just like how volumes live under a subject, and how solutions live under a session, runs should live under a session. There can be a runs.json that just lists run IDs (just like solutions.json and volumes.json do).

Define a Run class. In can live in the `plan/` directory

Database class can have similar methods to write new runs, overwrite rights, etc. Each Run should have the same basic serialization functionality: to/from file and to/from json. See how `Solution` is written and do a similar thing to that.

The attributes of a run: id (str), success flag (bool), note (str, large text field), associated session ID (str).
(We can fill in more attributes as we find that we need them)

Here is something extra we think we might want: the ability to save a snapshot of a session's protocol, transducer transform, and target alongside a run. Still need to think about how we want to do this...

ebrahimebrahim:
Closed via #131

PR #131, opened 2024-10-10
Add Runs #96

medmssaouri:


medmssaouri:
Hi @ebrahimebrahim, could you take an initial look at this commit. ( i haven't completed the test part yet, so please just ignore it for now), thanks

src/openlifu/db/database.py
```diff
@@ -107,6 +107,35 @@ def write_session(self, subject:Subject, session:Session, on_conflict=OnConflict
...
+        run.to_file(run_metadata_filepath)
+
+        # Write snapshot of the session
+        subject = self.load_subject(subject_id)
+        session = self.load_session(subject, session_id)
+        session.to_file(run_metadata_filepath.parent / f'{run.id}_session_snapshot.json')
+
+        # Write snapshot of the protocol
+        protocol = self.load_protocol(session.protocol_id)
+        protocol.to_file(run_metadata_filepath.parent / f'{run.id}_protocol_snapshot.json')
```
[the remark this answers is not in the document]
medmssaouri: I agree, in that case, i think it would be better to give only session as input with a bool argument write_snapshot. Protocol and Subject could be extracted from "session". For this approach, i want to make sure that even when the session is not saved yet, it points to a protocol and a subject that are saved ? or it could points to a subject and a protocol that are not saved yet.   
def write_run(self, session:Session, run:Run, write_snapshot=False, on_conflict=OnConflictOpts.ERROR):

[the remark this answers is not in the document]
medmssaouri: @ebrahimebrahim I agree with (1) and (2) but when the user does not provide a protocol as input of write_run, should we still write a snapshot of the session ? if yes, then should we write with it the protocol for session.protocol_id ?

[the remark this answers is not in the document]
ebrahimebrahim: I think it's the same with protocol: only write a protocol snapshot if the user provides it, and if the user does not provide it then don't write it.

medmssaouri:
@ebrahimebrahim Could you take a second look, thanks

src/openlifu/db/database.py
```diff
@@ -575,6 +626,13 @@ def get_volume_dir(self, subject_id, volume_id):
     def get_volume_metadata_filepath(self, subject_id, volume_id):
         return Path(self.get_volume_dir(subject_id, volume_id)) / f'{volume_id}.json'
 
+    def get_run_dir(self, subject_id, session_id, run_id):
+        run_dir = self.get_session_dir(subject_id, session_id) / 'runs' / f'{run_id}'
+        return run_dir
+
+    def get_run_metadata_filepath(self, subject_id, session_id, run_id):
```
ebrahimebrahim, line 633: (in the case of `Run`s it's not just metatdata but rather the full definition)

```suggestion
    def get_run_filepath(self, subject_id, session_id, run_id):
```

ebrahimebrahim (changes requested):
Looking good, a few changes requested and a question about how we should treat snapshots when overwriting.

src/openlifu/db/database.py
```diff
@@ -107,6 +107,34 @@ def write_session(self, subject:Subject, session:Session, on_conflict=OnConflict
...
+        Path(run_metadata_filepath).parent.parent.mkdir(exist_ok=True)
+        run.to_file(run_metadata_filepath)
+
+        if session:
+            # Write snapshot of the session
+            session.to_file(run_metadata_filepath.parent / f'{run.id}_session_snapshot.json')
+
+        if protocol:
+            # Write snapshot of the protocol
+            protocol.to_file(run_metadata_filepath.parent / f'{run.id}_protocol_snapshot.json')
```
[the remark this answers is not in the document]
medmssaouri: @ebrahimebrahim In my current code, a run with an id can only be written one time, and cannot be overwritten. Do you think that the user should be allowed to overwrite a run ?

src/openlifu/db/database.py
```diff
@@ -107,6 +107,34 @@ def write_session(self, subject:Subject, session:Session, on_conflict=OnConflict
...
 
+    def write_run(self, run:Run, session:Session = None, protocol:Protocol = None, on_conflict=OnConflictOpts.ERROR):
+        # Check whether the run already exist in the session
+        run_ids = self.get_run_ids(session.subject_id, session.id)
+
+        if run.id in run_ids:
+            if on_conflict == OnConflictOpts.ERROR:
+                raise ValueError(f"Run with ID {run.id} already exists for session {session.id}.")
+            elif on_conflict == OnConflictOpts.OVERWRITE:
+                raise ValueError("OVERWRITE is not supported for write_run")
```
ebrahimebrahim, line 118:
```suggestion
                raise ValueError("Runs are write-once objects and may not be overwritten.")
```

(just to make the intention clear -- it's not just not supported but it's deliberately disallowed)

src/openlifu/plan/__init__.py
```diff
@@ -1,7 +1,9 @@
 from .protocol import Protocol
+from .run import Run
```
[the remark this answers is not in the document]
medmssaouri: I think we can open another issue specifically for this for all other classes, because there is a problem of circular import that will probably need sometime to fix

src/openlifu/db/database.py
```diff
@@ -117,6 +117,48 @@ def write_session(self, subject:Subject, session:Session, on_conflict=OnConflict
 
         self.logger.info(f"Added session with ID {session_id} for subject {subject.id} to the database.")
 
+    def write_run(self, run:Run, session:Session = None, protocol:Protocol = None, on_conflict=OnConflictOpts.ERROR):
+        """Write a run with a snapshat of session and a snapshat of protocol if provided
```
ebrahimebrahim, line 121: mis-spelling: snapshot

ebrahimebrahim, line 121:
```suggestion
        """Write a run with a snapshot of session and a snapshot of protocol if provided
```

src/openlifu/db/database.py
```diff
@@ -117,6 +117,48 @@ def write_session(self, subject:Subject, session:Session, on_conflict=OnConflict
...
+
+        Args:
+            run: run to be written
```
ebrahimebrahim, line 124:
```suggestion
            run: Run to be written
```

src/openlifu/db/database.py
```diff
@@ -117,6 +117,48 @@ def write_session(self, subject:Subject, session:Session, on_conflict=OnConflict
...
+            session (optional): the session used in the run, it will have a snapshot written alongside the run
```
ebrahimebrahim, line 125:
```suggestion
            session (optional): the Session used in the Run, it will have a snapshot written alongside the run
```

src/openlifu/db/database.py
```diff
@@ -117,6 +117,48 @@ def write_session(self, subject:Subject, session:Session, on_conflict=OnConflict
...
+            protocol (optional): the protocol used in the run, it will have a snapshot written alongside the run
```
ebrahimebrahim, line 126:
```suggestion
            protocol (optional): the Protocol used in the Run, it will have a snapshot written alongside the run
```

ebrahimebrahim (dismissed):
Looks great, thank you!
I made some minor style suggestions. Once you fix those please go ahead and merge,

ebrahimebrahim (approved):

tests/test_database.py
```diff
@@ -95,6 +96,56 @@ def test_write_session(example_database: Database, example_subject: Subject):
...
+
+    # Make sure the Runs folder and Runs file are created
+    new_session = Session(id='new_session',subject_id='example_subject')
+    example_database.write_session(subject, new_session)
+    new_run = Run(id=run_id, success_flag=success_flag, note=note, session_id='new_session', solution_id=solution_id)
+    example_database.write_run(new_run, new_session, protocol, on_conflict=OnConflictOpts.OVERWRITE)
+    runs_filename = example_database.get_runs_filename(subject.id, new_session.id)
+    assert runs_filename.exists()
+    assert runs_filename.is_file()
+    assert runs_filename.name == "runs.json"
```
medmssaouri, line 133: @ebrahimebrahim  before i merge, i noticed that it could be useful to add this test that make sure that Runs file is created when a run is written for the first time for a particular session

PR #133, opened 2024-10-10
Fix write_volume to handle new subject directory

sadhana-r:


medmssaouri:
@sadhana-r Thanks, i agree with the bug fix, i will approve once the tests pass

src/openlifu/db/database.py
```diff
@@ -317,7 +318,7 @@ def get_volume_ids(self, subject_id):
             self.logger.info("Volume IDs for subject %s: %s", subject_id, volume_ids)
             return volume_ids
         else:
-            self.logger.warning("Volumes file not found for subject %s.", subject_id)
+            self.logger.info("Volumes file created for subject %s.", subject_id)
```
medmssaouri, line 321: @sadhana-r i think that we don't create a file here when it is not found ?

sadhana-r: The volumes.json file isn't created here but `write_volume_ids` creates the volumes.json file if the volume_ids returned by this function is empty. 
I didn't think this should be logged a warning since its okay for there to not be a volumes file found in the event that a new subject is being referened. Maybe I can keep the original message but change it from warning to info? And I could also add in the  `write_volume_ids`, a log message for when a new file is created. What do you think?

medmssaouri: Since get_volume_ids can be called alone, i think that we should keep the previous message with the modification from warning to info. And as you suggest, adding a log message to write_volume_ids when new file is created.
Thanks

medmssaouri (approved):
Thanks, LGTM !!

Issue #104, opened 2024-09-13
Port over minimal solution analysis code from matlab


ebrahimebrahim:
Port the [solution analysis code](https://github.com/OpenwaterHealth/opw_neuromod_sw/blob/1fe97a8245202c24cdcdc8ebabc147256c8da186/%2Bfus/%2Btreatment/Solution.m#L47) from matlab into openlifu, taking only what we need to take for now in order to get the pulse needed to ultimately tell the hardware what to do.

The rest of the solution analysis code will be handled in #105 later

Depends on #103.

ebrahimebrahim:
Re-opening after I accidentally merged prematurely. Reverted that merge.

PR #135, opened 2024-10-16
Port over minimal solution analysis code from matlab (#104)

ltetrel:
This adds minimal solution analysis code so it outputs at least the `mainlobe_pnp_MPa` attribute. The rest has been commented out for future work, additionnal modules/functions are also added to prepare future work.

Please check the #TODO comments in the code for technical discussions.


ltetrel:
@ebrahimebrahim thanks, was just about to do it,

src/openlifu/bf/calc_dist_from_focus.py
```diff
@@ -0,0 +1,42 @@
...
+from xarray import DataArray
+
+from openlifu.bf import offset_grid
+from openlifu.geo import Point
+
+
+def calc_dist_from_focus(
+        data_arr: DataArray,
+        focus: Point,
+        # units: str = "m", #TODO: I removed since input is already rescaled coords (already check were it is used)
```
ebrahimebrahim, line 13: :+1:

src/openlifu/plan/solution.py
```diff
@@ -64,14 +102,221 @@ class Solution:
...
+            # output_signal = np.zeros((transducer.numelements(), len(input_signal)))
+            # for i in range(transducer.numelements()):
+            #     apod_signal = input_signal * self.apodizations[focus_index, i]
+            #     output_signal[i] = transducer.elements[i].calc_output(apod_signal, dt)
+
+            # p0_Pa = np.max(output_signal)
+
+            # get focus region masks (for mainlobe, sidelobe and beamwidth)
+            mainlobe_mask = mask_focus(
+                self.simulation_result,  #TODO: Original code uses coords, but too complicated to maniplulate a Coordinates class
```
ebrahimebrahim, line 182: Can you elaborate on this comment -- what additional things do we need to do with coords

src/openlifu/plan/solution.py
```diff
@@ -64,14 +102,221 @@ class Solution:
...
+        # treatment_dutycycle = self.get_treatment_dutycycle()
+        # ita_mWcm2 = self.get_ita(output, units="W/cm^2")
+
+        # power_W = np.zeros(self.num_foci())
+        # TIC = np.zeros(self.num_foci())
+        for focus_index in range(self.num_foci()):
+            if isinstance(self.focus, list):
+                foc = self.focus[focus_index]
+            elif isinstance(self.focus, Point):
+                foc = self.focus
```
ebrahimebrahim, line 171: Since you are iterating over foci here already, I feel that we could merge #127 first before this so that this can rely on the list of foci always being an actual list. If you agree please go ahead and review #127

src/openlifu/plan/solution.py
```diff
@@ -64,14 +102,221 @@ class Solution:
...
+
+            # p0_Pa = np.max(output_signal)
+
+            # get focus region masks (for mainlobe, sidelobe and beamwidth)
+            mainlobe_mask = mask_focus(
+                self.simulation_result,  #TODO: Original code uses coords, but too complicated to maniplulate a Coordinates class
+                foc,
+                options.mainlobe_radius,
+                mask_op=MaskOp.LESS_EQUAL,
+                units="m",
```
ebrahimebrahim, line 186: These units "m" are the units of the `options.mainlobe_radius`, right? If so I think that the units really should be specified in that `options` object, or at least there should be a docstring under the `mainlobe_radius` attribute of the SolutionOptions class indicating that it must be specified in meters

ebrahimebrahim (commented):
Exciting to see this coming together!

I focused on the parts that are involved in getting `mainlobe_pnp_MPa` and left some broad comments for now

After another round of changes and after we discuss in our meeting soon you can request review again and I will dive into more detail. Perhaps we'll want to merge #127 first as well, as I mention below.

src/openlifu/bf/calc_dist_from_focus.py
```diff
@@ -0,0 +1,42 @@
+from typing import Tuple
+
+import numpy as np
+from xarray import DataArray
+
+from openlifu.bf import offset_grid
```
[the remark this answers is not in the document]
ltetrel: Thanks!

src/openlifu/plan/solution.py
```diff
@@ -64,14 +102,221 @@ class Solution:
...
+        if isinstance(self.focus, list):
+            nfoc = len(self.focus)
+        elif isinstance(self.focus, Point):
+            nfoc = 1
+        else:
+            raise ValueError("Cannot get number of foci for types other than Point.")
+
+        return nfoc
+
+    def analyse(self, transducer: Transducer, options: SolutionOptions = SolutionOptions()) -> SolutionAnalysis:
```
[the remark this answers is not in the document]
ltetrel: Actually they use `analyze` so I switched to that :D

tests/test_solution.py
```diff
@@ -55,15 +94,22 @@ def test_json_serialize_deserialize_solution(
...
+
+def test_save_load_solution_custom_dataset_filepath(example_solution: Solution, tmp_path: Path):
     """Test that a solution can be saved to and loaded from disk faithfully, this time with a custom path for simulation data."""
     json_filepath = tmp_path/"some_directory"/"example_soln.json"
     nc_filepath = tmp_path/"some_other_directory"/"sim_output.nc"
     example_solution.to_files(json_filepath, nc_filepath)
     assert dataclasses_are_equal(Solution.from_files(json_filepath, nc_filepath), example_solution)
+
+
+def test_solution_analysis(example_solution: Solution, example_transducer: Transducer):
```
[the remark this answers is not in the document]
ltetrel: Yep makes sense, do we want to have tests against a reference data, or only the run (like for `analyze)`?

src/openlifu/plan/solution.py
ltetrel: So in terms of pure functionality, only coordinates are needed at this step since masking require only spatial information (simulated sound pressure is not needed). However, I was not able to update only the `xarray.Coordinates` from the `xarray.DataArray`, hence the comment.

Howver in terms of architecture, it could make sense to have a child class (originally called `Volume` in matlab) derived from `xarray.DataArray` that would include members to perform all geometric ops like `get_ndgrid_from_arr` `rescale_coords` `rescale_data_arr` etc...

src/openlifu/plan/solution.py
ltetrel: I prefer to add the foci list feature later so we keep the minimal solution clean of this.

I added it because it was some leftovers from my initial development, but the rest is actually not really compatible with a foci list (few attributes from `solution_analysis` should be a list).

src/openlifu/plan/solution.py
ltetrel: From my understanding, `aspect_ratio` does not have units, it is just a scalar to scale and take into account voxel anisotropy. the `units` here are used to defined to what units all the spatial dimension should be rescaled to.

@peterhollender you coonfirm ?

ltetrel:
@ebrahimebrahim as suggested, there may be some regression in the solution analysis `mainlobe_pnp_MPa` that can be resolved later in another issue.

When you agree with the current status, I will squash everything so we have only one commit.

ebrahimebrahim:
Sounds good and thank you for all this work -- will review soon (later today or early Mon) and let you know

ebrahimebrahim (approved):
I tried it out, everything looks good as far as I can tell. Thank you for all the work on this and the unit tests. As we discussed you can go ahead and open an issue to track the fact that we do not yet know for sure that `analyze` is producing the correct output. I think the same of `get_beamwidth`, right? It is not unit tested and we don't know that it is working for sure, so you can mention that in the issue as well.

ebrahimebrahim:
OOPS.
I merged before checking that your commits follow the required style.
Sorry this is my fault, I think you said you would squash everything later, but I forgot about this.
I am going to undo this by ripping the commits out of the main branch (:shushing_face:),
then can you please re-make this PR and just squash the commits? TY and sorry about that

Issue #126, opened 2024-10-02
Port over solution creation and scaling code


ebrahimebrahim:
In #104 the focus was on porting over the [solution analysis function](https://github.com/OpenwaterHealth/opw_neuromod_sw/blob/1fe97a8245202c24cdcdc8ebabc147256c8da186/%2Bfus/%2Btreatment/Solution.m#L47).

In this issue we are indicating that we also need
- something like [`calc_solution`](https://github.com/OpenwaterHealth/opw_neuromod_sw/blob/1fe97a8245202c24cdcdc8ebabc147256c8da186/%2Bfus/%2Btreatment/Plan.m#L63) to act as a simulation runner. Note that `Plan` in matlab becomes `Protocol` in openlifu-python, and also note that there was an initial attempt to do something similar to `calc_solution` [here in SlicerOpenLIFU](https://github.com/OpenwaterHealth/SlicerOpenLIFU/blob/c4f783a86ce52a9a8246ac3befaf7ab227a464d5/OpenLIFUSonicationPlanner/OpenLIFUSonicationPlanner.py#L316).
- especially the [`scale_solution`](https://github.com/OpenwaterHealth/opw_neuromod_sw/blob/1fe97a8245202c24cdcdc8ebabc147256c8da186/%2Bfus/%2Btreatment/Plan.m#L269) aspect which uses part of the solution analysis to rescale the solution and `Pulse`. This is the criticial part that we need for a minimal workflow to be viable: to determine the amplitude of pulse 

Depends on #104 
Would be nice to have #125 done before starting on this

ebrahimebrahim:
When we port over `calc_solution` we will want to stack simulation outputs over foci.

In the initial attempt in SlicerOpenLIFU (linked above) you can see that it was done by `List[PlanFocus]` where a `PlanFocus` contained a simulation output xarray. We will get away from this, because of the fix introduced in #127 .

After the fix in #127 here is a way to stack multiple simulation outputs into the correct sort of xarray dataset.

```py
combined_dataset = xarray.concat(
    [
        sim.assign_coords(focal_point_index=i)
        for i,sim in enumerate(sims)
    ],
    dim='focal_point_index',
)
```

assuming `sims` is a list of individual xarray datasets (one for each focal point and ordered in the same order as the list of focal points) that are output of `openlifu.sim.run_similation`, and assuming that

PR #139, opened 2024-10-23
[do not merge] Temporary branch to test S#124

ebrahimebrahim:
This is a temporary branch created just to test https://github.com/OpenwaterHealth/SlicerOpenLIFU/pull/124 until the needed functionality is ported from the matlab codebase (#104 and #126)

This PR is meant to be used for reference and then closed without merging.

The important thing to reference here in this branch is `Protocol.calc_solution`. This shows the API that we will assume from SlicerOpenLIFU


ebrahimebrahim:
@ltetrel When you do #126 please reference `Protocol.calc_solution` in this temporary branch to see what is the API we need. Your implementation maybe be different because it would be following the matlab code more closely, but the API can hopefully be almost the same

ebrahimebrahim:
This is no longer needed since #146 is merged.

PR #143, opened 2024-10-26
DB create operations initialize empty component object lists

### LINKED ISSUE #142: Writing a solution to a newly created session produces a warning
When using `Database` to create a new session, it is not immediately possible to write a solution using `Database.write_solution` because there isn't a solutions.json file

Similar issues that I am going to lump into this one:
- when you make a session it doesn't create the runs.json
- when you make a subject it doesn't create the volumes.json
- when you make a subject it doesn't create the sessions.json

ebrahimebrahim:
* automatically create empty `runs.json`, `solutions.json`, and `volumes.json` when new objects are written to the database that should have those
* update our example data in dvc to have these empty solutions.json, etc. files

Close #142

### DIFF
--- a/db_dvc.dvc
+++ b/db_dvc.dvc
@@ -1,6 +1,6 @@
 outs:
-- md5: 272d36e8069e0b15bd61da82a8a4c5b9.dir
-  size: 21808282
-  nfiles: 36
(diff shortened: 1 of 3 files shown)


sadhana-r (approved):
Looks good!

PR #159, opened 2024-11-12
add mannequin example, fix some database write operations

### LINKED ISSUE #158: writing new protocols is not possible
- writing new protocols is not possible because the the folder in which to save the json file is not created
- when a protocol is written (e.g. via `Protocol.to_file`) it's not always reconstructed faithfully -- the `protocol.analysis_options.mainlobe_aspect_ratio` switches around between being a `list` or a `tuple`
- currently `Databse.write_protocol` does its own thing instead of using `Protocol.to_file`
- protocol writing functionality is not unit tested so we are not catching any of these issues

### LINKED ISSUE #160: overwriting a subject or session clears out IDs of child objects
Overwriting a subject clears out the session IDs in the sessions.json file, creating an invalid database where the sessions.json shows an empty list of IDs while the session file data is left sitting there inaccessible by `Database`.

Similarly, overwriting a session clears out the solution IDs and the run IDs in the solutions.json and runs.json files.

ebrahimebrahim:
- Close #158 
- Close #160
- Add mannequin example
- Standardize example database formatting

### DIFF
--- a/db_dvc.dvc
+++ b/db_dvc.dvc
@@ -1,6 +1,6 @@
 outs:
-- md5: b628b5613c9021c1023db5a448476136.dir
-  size: 21814345
-  nfiles: 48
+- md5: c4d9b9dba27c00aa19810dcab4672fd0.dir
+  size: 752385132
+  nfiles: 63
   hash: md5
   path: db_dvc
--- a/db_dvc.dvc
+++ b/db_dvc.dvc
@@ -1,6 +1,6 @@
 outs:
-- md5: b5ae75d0ad430394724f725775f97edf.dir
-  size: 752385282
-  nfiles: 65
+- md5: ea50713541a8d011e684501d03d3fa75.dir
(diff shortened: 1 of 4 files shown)


ebrahimebrahim:
Note the photoscans folder -- this a proposed way to organize the photogrammetry files

I was trying to think of a good name for the photogrammetry mesh+texture data.

It looks like openwater folks have been calling them just "scans" but that seems like it could get confused with MRI volumes. Sometimes they called them "phone scans" but openlifu-python doesn't care if the capture device is a phone.

So the best I can come up with is "photoscans" -- open to suggestions

ebrahimebrahim:
> In the example data, for the sessions in example_subject and mannequin, the solutions.json file in the solutions subfolder is missing the required solution_ids for the solutions in the folder.

:fearful: oh, glad you caught this!

sadhana-r (approved):

PR #162, opened 2024-11-14
 Port over a minimal virtual fitting approach from matlab (#147)

### LINKED ISSUE #147: Port over a minimal virtual fitting approach from matlab
The virtual fitting step is needed to get an initial position for the transducer. 

Given a segmented MRI from the head of a patient (outer surface of the skin) and a target focus point, we want to:
1. find the point where the trajectory intersects the skin surface
2. extract a local patch of skin surface positions around that point
3. fit those positions to a plane, with the intent of it being approximately tangent to the skin surface at that point
4. use the detected position and plan orientation to define a transformation that describes the local coordinates normal to the skin surface around that point
5. offset that transformation using the offset_matrix to get the expected position and orientation of the transducer when accounting for the standoff

With that algorithm in hand, we can define a grid of search trajectories (th, phi), and for each trajectory, calculate the estimated transducer position along the trajectory, and analyze that position. To analyze the position, minimally we need to detect the location of the target in the transducer coordinate system and check it against the transducer steering limits.

ltetrel:
This adds the minimal implementation for the virtual fitting algorithm.

Close #147


src/openlifu/vf/virtual_fit.py
```diff
@@ -0,0 +1,161 @@
...
+    """Search grid units."""
+
+    steering_limits: Tuple[TargetConstraints] = field(default_factory=list)
+    """Defines the accepteable range for a target in the transducer space, usually (lat, ele, ax)."""
+
+    blocked_elems_threshold: float = 0.1
+    """How much blocked elements are acceptable."""
+
+    volume: xa.Dataset = field(default_factory=xa.Dataset)
+    """The MRI volume on which to optimize the position."""
```
ebrahimebrahim, line 44: Just confirming, this volume is assumed to be already in transducer coordinates, correct?

src/openlifu/vf/virtual_fit.py
```diff
@@ -0,0 +1,161 @@
...
+    """The yaw range for the grid search."""
+
+    yaw_step: int = 3
+    """The yaw step for the grid search."""
+
+    search_range_units: str = "deg"
+    """Search grid units."""
+
+    steering_limits: Tuple[TargetConstraints] = field(default_factory=list)
+    """Defines the accepteable range for a target in the transducer space, usually (lat, ele, ax)."""
```
ebrahimebrahim, line 38: Should these properties, from `pitch_range` to `steering_limits`, be properties of the Transducer? Or of the Protocol? I feel like it's one of those... if agreed we can open an issue for later putting them in Transducer or Protocol. @peterhollender any opinion?

peterhollender: pitch and yaw ranges should come from the Protocol, as they are determined by the target's general location

src/openlifu/vf/virtual_fit.py
```diff
@@ -0,0 +1,161 @@
...
+        #     except ValueError:
+        #         pos_analysis = 0.0
+        #
+        # return pos_analysis
+
+        pass
+
+    def run(
+            self,
+            target: Point,
```
ebrahimebrahim, line 119: What is assumed about the coordinates of this target?

E.g. is it in the same coordinates as the indices of the volume array?

Or is it in the lat, ele, ax coordinates?

src/openlifu/vf/virtual_fit.py
ebrahimebrahim: Makes sense. And steering limits are more of a Transducer thing right? Like we would do some simulations to determine steering limits for each Transducer and then hardcode those limits into the Transducer definition

src/openlifu/vf/virtual_fit.py
ebrahimebrahim: Actually no... I do still have this question... is the volume meant to be in lat, ele, ax coords here?

ebrahimebrahim: Sorry I keep confusing myself. I think almost certainly the answer is yes: the volume is in lat, ele, ax coordinates, and so is the target.

src/openlifu/vf/virtual_fit.py
peterhollender: Target is in LPS coordinates (coordinates of the volume)

src/openlifu/vf/virtual_fit.py
```diff
@@ -0,0 +1,161 @@
...
+                yaw, pitch = (search_grid[0][i, j], search_grid[1][i, j])
+                self.logger.info(f"VirtualFit: Analysing {(yaw, pitch)}...")
+                # 3. define transducer transform (plane fitting) on the surface (skin) given spherical coordinate (yaw, pitch)
+                # self.fit_to_surface(sph_coords: Tuple[float, float], skin_surface: np.ndarray)
+                # 4. analyse current transform
+                # self.analyse_position(pos: np.ndarray, transducer: Transducer, target: Point):
+                optimal_transform = np.zeros((4, 4))
+        self.logger.info("VirtualFit: Found optimal position!")
+
+        return optimal_transform
```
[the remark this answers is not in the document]
peterhollender: I'd assume Volume units, typically mm. In fact, the only significant use of changing units is for setting up the k-wave sims, which have to be m. Otherwise, I'd have just forced it all into mm.

src/openlifu/vf/virtual_fit.py
ltetrel: It is indeed in (lat, ele, ax), if we don't satisfy this we would need to resample to that coordinate before virtual fitting (as for `Protocol.calc_solution`). Just updated the docstrings in a new commit. https://github.com/OpenwaterHealth/OpenLIFU-python/pull/162/commits/529404002778fe2ccaa028c3b67a51f3d17cee03

src/openlifu/vf/virtual_fit.py
ltetrel: If that is the case, then we would probably remove those attributes from `VirtualFit`, we don't want to duplicate those on disk (the json file associated to `VirtualFit`). For now I will leave them for simplicity, and we can sync about that before merging.

src/openlifu/vf/virtual_fit.py
ltetrel: In `Protocol.calc_solution` we assumed the target to be in same coordinates as the volume. Because Slicer resample to (lat, ele, ax), the target is also in that. Because we did not implement the resampling routine in the python lib we should also assume (lat, ele, ax) here for short term.

src/openlifu/vf/virtual_fit.py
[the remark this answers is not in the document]
ltetrel: Also the transform unit (and tr matrix) is defined in`array_transform` from `subject/sessions/session.json`. So we can convert it later. Just updated the `optimal_transform` typing for clarity. https://github.com/OpenwaterHealth/OpenLIFU-python/pull/162/commits/5594c67057ff754860b698b76149bdcb0bbfa087

ltetrel:
Current state for this PR:
* @TongFuKitware was able to develop the functions necessary for this workflow to run, it was tested by loading output data from the old matlab code (test code + data sent to @ebrahimebrahim).
* we have the `VirtualFit` class skeleton in place, with its members callable as standalone functions.
* there are still discussions about which attributes for the class, and eventually some to put in other classes (ex. transducer stand-off geometry in `Transducer` class).
* `extract_skin_surface` is called in the `__post_init__` function and stores the segmented skin (np.ndarray) in a new attribute, as well as the fitted interpolator. This to be able to call `VirtualFit.run` with different targets and not re-segmenting the volume.
* a mesh can be constructed as well to replace the interpolator and segmented skin (list of points as np.ndarray), and benefit from normals (which can help during transducer pose search).
* we are still missing proper end-to-end workflow (interaction with Slicer scene)
* we are still missing tests with expected data

ebrahimebrahim:
@peterhollender Quick question, in the manual skin surface extraction protocol. after getting the skin surface how did you determine an appropriate skin surface _center point_? i.e. the origin of the spherical coordinate system in which the virtual fitting search takes place.

peterhollender:
> @peterhollender Quick question, in the manual skin surface extraction protocol after getting the skin surface how did you determine an appropriate skin surface _center point_? i.e. the origin of the spherical coordinate system in which the virtual fitting search takes place.

In the existing Matlab implantation, I used the location of the target as the search origin. Of course, in that initial work, the target was located anterior in the brain and near the L-R centerline, so the forehead formed a natural curve around it. I guess for other targets, you might rotate the coordinate system to put the angles more in line with the expected approach- I'd rotate about the `IS` axis, and then the `LR'` axis

ebrahimebrahim:
Major changes:
- The input volume is now just given as a numpy array with an affine,  not an xarray thing. This is more like how a user would actually  have a volume after reading it from disk.
- Many operations are now vectorized rather than relying on a python loop.   There is still a python loop over the virtual fit search grid points,  but there is no inner python loop over the plane-fitting points.
- We work in RAS rather than LPS, since typically when you load a nifti  file into memory it will come in with an affine that maps into RAS.  We can always add a conversion layer later if we want to add flexibility  with how the volume and target are provided. (The transducer transforms  that are returned however still map into LPS space, since all openlifu  transducer transforms have been working this way.) This is just keeping  things simpler for the initial approach.
- The algorithm has been written in an imperative functional style rather than using  a dataclass. This is easier to read and iterate on at first, and while  some object oriented approach probably makes sense eventually, it is  hard to get it exactly right without first having a fully working  algorithm in place to see how it fits in with the rest of the  application.
- The virtual_fit module has been moved to the top level, so no more vf  package to go through first. We can move it into its own vf package if  we add tons of new independent files, but to start with we should
 organize things more simply.
- (internal change) yaw and pitch are internally converted to spherical  coordinates $\theta$ and $\phi$ to make it easier to work in that  coordinate system. Also $\theta$ and $\phi$ were swapped from the minor  previously existing usage, to match the physics convention. The relation  is that pitch is $\phi$ and yaw is $90-\theta$. The $90-\theta$ is there  to allow us to use a more standard version of spherical coordinates  (where previously we had a confusing swap of sines with cosines from the  standard approach).

ebrahimebrahim (approved):

PR #171, opened 2024-11-25
Fix matrix array generation

### LINKED ISSUE #170: gen_matrix_array does not center array
The code used in gen_matrix_array does not ensure that the array is centered at 0,0,0, because it uses the `//` operator instead of the `/`.

peterhollender:
Closes #170

### DIFF
--- a/src/openlifu/xdc/transducer.py
+++ b/src/openlifu/xdc/transducer.py
@@ -203,8 +203,8 @@ class Transducer:
     @staticmethod
     def gen_matrix_array(nx=2, ny=2, pitch=1, kerf=0, units="mm", impulse_response=1, impulse_dt=1, id='array', name='Array', attrs={}):
         N = nx * ny
-        xpos = [(i - nx // 2) * pitch for i in range(nx)]
-        ypos = [(i - ny // 2) * pitch for i in range(ny)]
+        xpos = [(i+0.5 - nx / 2) * pitch for i in range(nx)]
+        ypos = [(i+0.5 - ny / 2) * pitch for i in range(ny)]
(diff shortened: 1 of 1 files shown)


src/openlifu/xdc/transducer.py
```diff
@@ -203,8 +203,8 @@ def to_json(self, compact:bool) -> str:
     @staticmethod
     def gen_matrix_array(nx=2, ny=2, pitch=1, kerf=0, units="mm", impulse_response=1, impulse_dt=1, id='array', name='Array', attrs={}):
         N = nx * ny
-        xpos = [(i - nx // 2) * pitch for i in range(nx)]
-        ypos = [(i - ny // 2) * pitch for i in range(ny)]
+        xpos = [(i+0.5 - nx / 2) * pitch for i in range(nx)]
+        ypos = [(i+0.5 - ny / 2) * pitch for i in range(ny)]
```
sadhana-r, line 207: This works well.  It would be helpful if you could you add some docstrings to this function to explain what the different arguments are and that this function is meant to create an array centered at (0,0,0). 
An alternative suggestion for getting a centered array that avoids these operations is
`start = -0.5`
`stop = 0.5`
`xpos = np.linspace(start,stop,nx)*pitch`
`ypos = np.linspace(-start,stop,ny)*pitch`
If the 0.5 spacing between the coordinates is important, you can scale the start/stop by `(nx-1)`. Just a thought but you don't have to change it.

sadhana-r (changes requested):

sadhana-r (approved):

Issue #175, opened 2024-11-26
protocol.calc_solution does not use GPU


peterhollender:
`kwave.kspaceFirstOrder3D` has CPU (C++) and GPU (CUDA)-accelerated binaries, with the CUDA version providing substantial performance improvements over the CPU version. However, in the current implementation of `protocol.calc_solution`, the call to `run_simulation` is hard coded with `gpu = False`, which disables the use of the CUDA version and significantly degrades performance. 

In the MATLAB version of the code, I would have the simulation code _try_ to use the CUDA code and then fall back to the C++ code if the GPU resource wasn't available (via an exception), but we may want a cleaner way to check if the GPU is available here up front. In fact, one of our product requirements is to verify that sufficient computation resources are available, so we could have it always use the GPU and through a well-formed error if the GPU isn't available, although I would prefer that we take more of a "stern warning" approach so that CPU-bound users aren't completely shut out, but clinical system operators using the custom application would be required to be using the appropriate hardware.

ltetrel:
We could either check with https://github.com/NVIDIA/cuda-python or manually with 
```python
import subprocess

try:
    subprocess.check_output("nvidia-smi")
except Exception:
    print("No Nvidia GPU detected.")
```

PR #176, opened 2024-11-26
Default to using the GPU

peterhollender:
This addresses #175, but we still need a clean way to fall back to the CPU if the GPU isn't available.

### DIFF
--- a/src/openlifu/util/checkgpu.py
+++ b/src/openlifu/util/checkgpu.py
@@ -1,0 +1,12 @@
+from pynvml import nvmlDeviceGetCount, nvmlInit, nvmlShutdown
+
+
+def gpu_available() -> bool:
+    """Check the system for an nvidia gpu and return whether one is available."""
+    try:
+        nvmlInit()
+        device_count = nvmlDeviceGetCount()
+        nvmlShutdown()
+        return device_count > 0
+    except Exception: # exception could occur if there is a driver issue, for example
+        return False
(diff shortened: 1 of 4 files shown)


ebrahimebrahim:
Thanks for this. I may add some commits directly to this branch soon, to fall back to CPU.

ebrahimebrahim:
@peterhollender Can you have a look? It's your PR so I can't have you review it. I will add the approval if you verbally approve. I will add a comment to the part i wnat you to look at

src/openlifu/util/checkgpu.py
```diff
@@ -0,0 +1,12 @@
+from pynvml import nvmlDeviceGetCount, nvmlInit, nvmlShutdown
+
+
+def gpu_available() -> bool:
+    """Check the system for an nvidia gpu and return whether one is available."""
+    try:
+        nvmlInit()
+        device_count = nvmlDeviceGetCount()
+        nvmlShutdown()
+        return device_count > 0
```
ebrahimebrahim, line 10: This is the part to focus on -- the method of checking for GPU availalability here is to check at the device level whether there is an nvidia device.

So this is not checking for example whether the CUDA toolkit is installed.

pynvml is super lightweight of a dependency compared to things that bring in cuda, so that's why I defaulted to that, but please let me know if the relevant thing to check for the purpose of k-wave gpu usage is actually the presence of CUDA

peterhollender: yeah, I think this is fine. _Maybe_ we'd want helpful messaging if the CUDA one tries to run and fails, but it could be a pain to test (we'd need a computer with an NVIDIA-but-not-CUDA GPU...)

ebrahimebrahim (approved):

PR #178, opened 2024-12-03
Add concept of photoscans

### LINKED ISSUE #177: Define photoscans
Add to openlifu the concept of a photoscan and the ability to serialize them to and from json. Photoscans are added to a particular session, similar to how volumes are added to a subject. Similar to how volumes metadata include the filepath to the mri file, photoscan metadata will include the filepath to the model file, mtl file, and texture file. It will also include an approval flag with indicates whether the scan is good enough for use in transducer tracking. This issue will add a method to the Database class to write/load a photoscan for a particular session. The Database should also be able to retrieve photoscans by ID.

This assumes that the photoscan <> mri transform is saved to sessions file and that a single photoscan won't be used across different sessions.

sadhana-r:
Closes #177 

I decided to include the mtl file as an optional input when creating a photoscan.


ebrahimebrahim:
@sadhana-r as part of this PR can you also add to db_dvc the exr file that peter shared in google drive over email? just the exr file, I think it will work with the existing obj file that is already in the mannequin demo_4_simulated session.

src/openlifu/db/database.py
```diff
@@ -478,6 +534,19 @@ def get_volume_info(self, subject_id, volume_id):
                     "name": volume["name"],\
                     "data_abspath": Path(volume_metadata_filepath).parent/volume["data_filename"]}
 
+    def get_photoscan_info(self, subject_id, session_id, photoscan_id):
+        photoscan_metadata_filepath = self.get_photoscan_metadata_filepath(subject_id, session_id, photoscan_id)
+        with open(photoscan_metadata_filepath) as f:
+            photoscan = json.load(f)
```
ebrahimebrahim, line 540:
```suggestion
        photoscan = json.loads(photoscan_metadata_filepath.read_text())
```
(unimportant style tip)
[Path.read_text](https://docs.python.org/3/library/pathlib.html#pathlib.Path.read_text) is super handy and this saves a level of indentation

tests/test_database.py
```diff
@@ -399,3 +399,49 @@ def test_write_solution_new_session(example_database:Database, example_subject:S
...
+    photoscan_name = "EXAMPLE_PHOTOSCAN_2"
+    model_data_path = Path(tmp_path/'test_db_files/example_photoscan_2.obj')
+    model_data_path.parent.mkdir(parents=True, exist_ok=True)
+    model_data_path.touch()
+    texture_data_path = Path(tmp_path/'test_db_files/example_photoscan_texture_2.exr')
+    texture_data_path.parent.mkdir(parents=True, exist_ok=True)
+    texture_data_path.touch()
+    example_database.write_photoscan(example_session, photoscan_id, photoscan_name, model_data_path, texture_data_path)
+
+    assert(example_database.get_photoscan_ids("example_subject", "example_session") == ["example_photoscan", "example_photoscan_2"])
```
ebrahimebrahim, line 429: Here it's better to test that the list has length 2 and contains both items -- they don't have to be in that order. That way the test is less brittle. (a brittle test is one that can complain even when nothing is wrong)

tests/test_database.py
```diff
@@ -399,3 +399,49 @@ def test_write_solution_new_session(example_database:Database, example_subject:S
...
     example_database.write_solution(session, solution)
+
+def test_get_photoscan_info(example_database:Database, tmp_path:Path): # TODO: Need to finish this!
+    subject_id = "example_subject"
+    session_id = "example_session"
+    photoscan_id = "example_photoscan"
+    photoscan_info = example_database.get_photoscan_info(subject_id, session_id, photoscan_id)
+    assert(photoscan_info["id"] == "example_photoscan")
+    assert(photoscan_info["name"] == "ExamplePhotoscan")
+    assert(Path(photoscan_info["model_abspath"]) == \
```
ebrahimebrahim, line 410: Instead of asserting that the abspath is exactly this, we can just assert that it points to an existing file.
That way if we ever restructure where the data files are stored the test doesn't have to break

src/openlifu/db/database.py
```diff
@@ -274,6 +276,50 @@ def write_volume(self, subject_id, volume_id, volume_name, volume_data_filepath,
 
         self.logger.info(f"Added volume with ID {volume_id} for subject {subject_id} to the database.")
 
+    def write_photoscan(self, session: Session, photoscan_id, photoscan_name, model_data_filepath, texture_data_filepath, mtl_data_filepath: Optional[Path] = None, on_conflict=OnConflictOpts.ERROR):
+        if not Path(model_data_filepath).exists():
+            raise ValueError(f'Model data filepath does not exist: {model_data_filepath}')
+        if not Path(texture_data_filepath).exists():
+            raise ValueError(f'Texture data filepath does not exist: {texture_data_filepath}')
```
ebrahimebrahim, line 283: You [can test](https://docs.pytest.org/en/7.1.x/how-to/assert.html#assertions-about-expected-exceptions) that these are working in your unit tests by doing
```py
with pytest.raises(ValueError, match="data filepath does not"):
  ...
```
If the code in the ... does not raise a ValueError with the `match` text in its message, then the test fails.

Unrelated: also a `FileNotFoundError` may be more suitable than `ValueError` here

src/openlifu/db/database.py
```diff
@@ -274,6 +276,50 @@ def write_volume(self, subject_id, volume_id, volume_name, volume_data_filepath,
...
+                self.logger.info(f"Skipping photoscan with ID {photoscan_id} for session {session.id} as it already exists.")
+                return
+            else:
+                raise ValueError("Invalid 'on_conflict' option. Use 'error', 'overwrite', or 'skip'.")
+
+        # Create photoscan metadata
+        photoscan_metadata_dict = {"id": photoscan_id, "name": photoscan_name, "model_filename": Path(model_data_filepath).name, "texture_filename": Path(texture_data_filepath).name, "photoscan_approved": False}
+        if mtl_data_filepath:
+            if not Path(mtl_data_filepath).exists():
+                raise ValueError(f'MTL filepath does not exist: {mtl_data_filepath}')
```
ebrahimebrahim, line 301: same as other comment

tests/test_database.py
```diff
@@ -399,3 +399,49 @@ def test_write_solution_new_session(example_database:Database, example_subject:S
...
+    assert(example_database.get_photoscan_ids("example_subject", "example_session") == ["example_photoscan", "example_photoscan_2"])
+
+    photoscan_filepath = example_database.get_photoscan_metadata_filepath("example_subject","example_session","example_photoscan_2")
+    assert(photoscan_filepath.name == "example_photoscan_2.json")
+    assert((photoscan_filepath.parent/"example_photoscan_2.obj").exists())
+    assert((photoscan_filepath.parent/"example_photoscan_texture_2.exr").exists())
+
+    # When writing to a new subject and session
+    subject = Subject(id="bleh",name="Deb Jectson")
+    example_database.write_subject(subject, on_conflict=OnConflictOpts.OVERWRITE)
```
ebrahimebrahim, line 438: The OVERWRITE isn't needed here, and shouldn't be here in fact because we want this to be new

ebrahimebrahim (changes requested):
Very clean work, a few minor changes requested. The style things you can freely ignore or take as you wish, but some other minor things are there.

---

BTW since you are now running the tests locally here is another fun thing. Instead of just running `pytest` you can run `pytest --cov-report=html --cov`. This will generate a folder of HTML files, then you can open the index.html in a web browser and browse through the code looking at a report of which lines of code get run during a test and which are still missed. Better not to obsess over it too much, but sometimes it is satisfying to try and hit every line or try and get a high coverage percent

ebrahimebrahim (approved):
looks good!

PR #183, opened 2024-12-05
Fix focus masking for solution analysis

### LINKED ISSUE #141: Regressions with minimal solution analysis
After testing the functionality from https://github.com/OpenwaterHealth/OpenLIFU-python/pull/140 and comparing with the expected solution analysis output from Matlab (only `mainlobe_pnp_MPa` ATM) , there are some regressions. This can come from two sources:
1. Simulation output: there are some differences in the k-wave simulations output between the matlab and python generated-one.
2. [Incorrect mask computation](https://github.com/OpenwaterHealth/OpenLIFU-python/blob/main/src/openlifu/bf/offset_grid.py): and more specifically when we define the mask in the transducer reference, This depends on [geo.py](https://github.com/OpenwaterHealth/OpenLIFU-python/blob/17a6983465b6021adca059ef53839d609aa43c56/src/openlifu/geo.py#L38) which was not tested previously.

`tests/test_mask_focus.py` and `tests/test_offset_grid.py` would need to be updated accordingly.

peterhollender:
Closes #141


ebrahimebrahim (approved):
All looks great, ty!

To standardize the commit message format I will squash-and-merge -- hope that's okay.

PR #186, opened 2024-12-12
Add functionality to load photoscans

### LINKED ISSUE #180: Add photoscan loading functionality
Now that we have a concept of photoscans in the openlifu Database, let's add a method of loading the photoscan data into in-memory data structures so that openlifu-python can provide a common format for the 3D registration library and for SlicerOpenLIFU to work with.

This functionality should not be part of Database, but maybe somehting like `openlifu.io`? Maybe it's own module like `openlifu.io.photoscans`?

The goal is to have some functions that make it easy to take an openlifu `Photoscan` object and load the textured mesh into memory as some data structure (tuple, dict, custom class, whatver) that consists of a vtkPolyData and an image array of some kind (a numpy array? a vtkTexture?)

Design considerations include the following downstream applications:
- SlicerOpenLIFU will use this to load the photoscan into memory and then it will construct a textured model node ([S#147](https://github.com/OpenwaterHealth/SlicerOpenLIFU/issues/147)) . We know how to construct a textured model node from [Sam's branch here](https://github.com/OpenwaterHealth/SlicerOpenLIFU/blob/texture-model/OpenLIFUTextureModel/OpenLIFUTextureModel.py)
- [Kedar's library](https://github.com/OpenwaterHealth/OpenLIFU-3DRegistration) should interoperate with this format as well
-  Need to be able to convert between an openlifu photoscan object and the dictionary/json format so that the photocan "approval" status can be updated in memory and written to file. 

Depends on #177

### LINKED ISSUE #185: Update input arguments for writing photoscans to database
When writing photoscans to the database, the current implementation takes an openlifu `Session` object as input. The session object is used to derive `Session.subject_id` and `Session.id` required for writing to database. Instead, it would be better if this function took the `subject_id `and `session_id` directly as input arguments, so that the subject and session objects don't neccesarily need to be loaded/created before adding a photoscan to the database. 

Since an openlifu Photoscan object is being created, this function can also take the photoscan object as input instead of the individual parameters. This will make it easier when the photoscan needs to be re-written when the approval status is updated by the user. 

Update: The photoscan object only includes the relative filepaths to the data. Therefore, when a new photoscan is being written to file, the absolute filepaths need to be specified in order to copy the files. These can be specified as optional arguments to the write_photoscan function. `write_photoscan` will update the photoscan object with the relative filepaths. 
In the case where the photoscan is being re-written (i.e. to update the approval status), only the photoscan object can be provided. Since the data files aren't specified, `write_photoscan` should include a check to make sure that the required data has been previously associated with the photoscan object.

sadhana-r:
Closes #180 
Closes #185

- [x] Add unit tests
- [x] Update OpenLIFU to follow slicer conventions. OpenLIFU will read and write meshes assuming the LPS convention and internally use the RAS convention for storing meshes.
- [x] Update example data - include initialized photoscans folders
- [x] Separate the loading of the model and texture files from the Photoscan class definition and remove the `model` and `texture` attributes. The Photoscan class should be kept light weight so it is easily serializable.
- [x] Move the photoscan functionality to a different folder. Doesn't make sense for it to be in the database folder.


src/openlifu/db/photoscan.py
```diff
@@ -0,0 +1,229 @@
...
+
+def load_texture(file_name):
+    if not Path(file_name).exists():
+        raise FileNotFoundError(f'Texture data filepath does not exist: {file_name}')
+    texture = read_as_vtkimagedata(file_name)
+    return texture
+
+def read_as_vtkpolydata(file_name):
+
+    valid_suffixes = ['.g', '.obj', '.stl', '.ply', '.vtk', '.vtp']
```
ebrahimebrahim, line 129: A more pretty approach could be to replace `valid_suffixes` by 

```
suffix_to_reader_dict = {
    '.ply' : vtk.vtkPLYReader,
    '.vtp' : vtk.vtkXMLPolyDataReader,
    # etc.
}
```

and then you can be like `reader = suffix_to_reader_dict[ext]()` instead having so many `elif`s

src/openlifu/db/photoscan.py
```diff
@@ -0,0 +1,229 @@
...
+        return poly_data
+
+def convert_numpy_to_vtkimage(image_numpy):
+
+    vtkimage_data = vtk.vtkImageData()
+    vtkimage_data.SetDimensions(image_numpy.shape[1], image_numpy.shape[0], 1)
+    vtkimage_data.SetNumberOfScalarComponents(image_numpy.shape[2], vtkimage_data.GetInformation())
+    pd = vtkimage_data.GetPointData()
+    new_rgb_data = image_numpy.reshape((-1, image_numpy.shape[2]))
+    #new_rgb_data = np.flipud(new_rgb_data) # To look like blender format upon loading. Image is instead flipped in texture module.
```
ebrahimebrahim, line 176: blender format?

src/openlifu/db/photoscan.py
```diff
@@ -0,0 +1,229 @@
...
+            poly_data = reader.GetOutput()
+        elif ext == ".g":
+            reader = vtk.vtkBYUReader()
+            reader.SetGeometryFileName(file_name)
+            reader.Update()
+            poly_data = reader.GetOutput()
+
+        return poly_data
+
+def convert_numpy_to_vtkimage(image_numpy):
```
ebrahimebrahim, line 169: This function could use a docstring and type annotations. What is the expected shape of `image_numpy`? it's only for 2D images right?

I feel this function could use a unit test. The test could create a tiny 2D numpy array with some known values, run it through this function to get a vtk image, and then just make sure that the two agree at some location where the value is known

src/openlifu/db/photoscan.py
```diff
@@ -0,0 +1,229 @@
...
+                # Combine channels into a single RGB image (H x W x 3)
+                rgb_data = np.stack([R, G, B], axis=-1)
+                rgb_data = rgb_data.astype(np.float32)
+                # Normalize the data to 0-255 range for compatibility with VTK
+                rgb_data = np.clip(rgb_data*(2**16-1), 0, 65535)
+                image_data = convert_numpy_to_vtkimage(rgb_data)
+
+        return image_data
+
+def convert_between_ras_and_lps(mesh):
```
ebrahimebrahim, line 219: this function needs type annotations and should be unit tested

ebrahimebrahim (changes requested):
Nice work, some changes/additions requested where I spotted things.

For where to put the photoscan.py module, I think since it doesn't belong in any of the current folders you can put at at the top level, similar to `geo.py`. If it ever gets a friend that belongs with it in the future then we can create a new folder for them. What do you think of that?

src/openlifu/db/photoscan.py
sadhana-r: I'll remove this line. The texture image needs to be flipped before it is applied to the model. When testing this, I was comparing this functionality to the the image when loaded using blender. But Sam's code for applying the texture in Slicer includes the image flip so the image does not need to get flipped here.

ebrahimebrahim (approved):
Great work and I love the new unit tests -- thank you!

PR #188, opened 2025-01-08
Update transducer tracking definition

### LINKED ISSUE #181: Add transducer registration surface and body to transducers
The **transducer registration surface** is a model of an open surface to be used for registration. In practice it’s probably a sub-mesh of a larger **transducer body** model, which is a closed surface meant for visualization of the transducer.

In openlifu-python, transducers will gain two additional but optional data objects: a transducer registration surface model model and a transducer body model. For now, from the point of view of openlifu-python, these are just a mysterious data files. Similar to how volumes are currently handled – you can load the volume metadata json but it will only give you a path to a datafile and not any loaded data. So:

- Add attributes to the Transducer class to optionally specify these accompanying data files -- these are simply path strings.
- The path strings are relative paths (so  just filenames, because the data can be stored alongside the transducer json in the same folder).
- Add a function to Database to fetch the absolute paths to these datafiles so that the user can decide how to load them and what to do with them.
- Add functionality to the database transducer writing function to provide the data files, which just does a copy (similar to how adding a volume to a subject works in Database).
- Unit-test all this, as always. And add an example to our db_dvc if possible

sadhana-r:
Closes #181

### DIFF
--- a/db_dvc.dvc
+++ b/db_dvc.dvc
@@ -1,6 +1,6 @@
 outs:
-- md5: 2be55533d9c831968105871a54383497.dir
-  size: 802948645
-  nfiles: 75
+- md5: 1481ca23ef375f16d53a4058fd859a87.dir
+  size: 803476587
+  nfiles: 77
   hash: md5
   path: db_dvc
--- a/src/openlifu/db/database.py
+++ b/src/openlifu/db/database.py
@@ -574,6 +574,22 @@ class Database:
         photoscan = Photoscan.from_file(photoscan_metadata_filepath)
         return photoscan
 
+    def get_transducer_info(self, transducer_id):
+        transducer_metadata_filepath = self.get_transducer_filename(transducer_id)
+        with open(transducer_metadata_filepath) as f:
+            transducer = json.load(f)
+            transducer_dict = {"id": transducer["id"],\
+                    "name": transducer["name"],\
+                    "elements": transducer["elements"], \
+                    "frequency": transducer["frequency"], \
+                    "units": transducer["units"], \
+                    "attrs": transducer["attrs"]}
+            if "registration_surface_model" in transducer:
+                transducer_dict["registration_surface_abspath"] = Path(transducer_metadata_filepath).parent/transducer["registration_surface_filename"]
+            if "transducer_body_model" in transducer:
+                transducer_dict["transducer_body_abspath"] = Path(transducer_metadata_filepath).parent/transducer["transducer_body_filename"]
+            return transducer_dict
+
     def load_standoff(self, transducer_id, standoff_id="standoff"):
         raise NotImplementedError("Standoff is not yet implemented")
         standoff_filename = self.get_standoff_filename(transducer_id, standoff_id)
--- a/src/openlifu/db/database.py
+++ b/src/openlifu/db/database.py
@@ -659,7 +659,6 @@ class Database:
         Returns:
             Corresponding Transducer object
         """
-        from openlifu.xdc import Transducer
         transducer_filename = self.get_transducer_filename(transducer_id)
         transducer = Transducer.from_file(transducer_filename)
         return transducer
--- a/src/openlifu/db/database.py
+++ b/src/openlifu/db/database.py
@@ -194,7 +194,7 @@ class Database:
 
         self.logger.info(f"Added subject with ID {subject_id} to the database.")
 
-    def write_transducer(self, transducer, on_conflict: OnConflictOpts=OnConflictOpts.ERROR):
+    def write_transducer(self, transducer, registration_surface_model_filepath: Optional[str] = None, transducer_body_model_filepath: Optional[str] = None, on_conflict: OnConflictOpts=OnConflictOpts.ERROR):
         transducer_id = transducer.id
         transducer_ids = self.get_transducer_ids()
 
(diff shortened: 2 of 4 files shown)


db_dvc.dvc
```diff
@@ -1,6 +1,6 @@
 outs:
-- md5: 964978fc7c5a135c71de8e2ea547e0e4.dir
-  size: 802948419
-  nfiles: 68
+- md5: 8ffdeee8e4f084b780c86f5faef5c002.dir
+  size: 803476509
+  nfiles: 77
```
ebrahimebrahim, line 4: When I do a `dvc pull` in this branch I get

```
Collecting                                                                                    |0.00 [00:00,    ?entry/s]
WARNING: Some of the cache files do not exist neither locally nor on remote. Missing cache files:
md5: 8ffdeee8e4f084b780c86f5faef5c002.dir                                                                               
Fetching
Building workspace index                                                                      |126 [00:00, 6.19kentry/s]
Comparing indexes                                                                              |126 [00:13, 9.51entry/s]
Applying changes                                                                              |0.00 [00:00,     ?file/s]
Everything is up to date.
ERROR: failed to pull data from the cloud - Checkout failed for following targets:
db_dvc
Is your cache up to date?
<https://error.dvc.org/missing-files>
```

Could it be an issue with the way you pushed some additional files? (Maybe you did `dvc add` but not `dvc push`?)

src/openlifu/xdc/transducer.py
```diff
@@ -21,6 +21,10 @@ class Transducer:
     frequency: float = 400.6e3
     units: str = "m"
     attrs: Dict[str, Any] = field(default_factory= dict)
+    registration_surface_filename: Optional[str] = ""
+    """Relative path to an open surface of the transducer to be used for registration"""
+    transducer_body_filename: Optional[str] = ""
+    """Relative path to the closed surface mesh for visualizing the transducer body"""
```
ebrahimebrahim, line 27: if it's `Optional[str]` maybe the default value ought to be `None` instead of `""`

src/openlifu/db/database.py
```diff
@@ -192,7 +192,7 @@ def write_subject(self, subject, on_conflict=OnConflictOpts.ERROR):
 
         self.logger.info(f"Added subject with ID {subject_id} to the database.")
 
-    def write_transducer(self, transducer, on_conflict: OnConflictOpts=OnConflictOpts.ERROR):
+    def write_transducer(self, transducer, registration_surface_model_filepath: Optional[str] = None, transducer_body_model_filepath: Optional[str] = None, on_conflict: OnConflictOpts=OnConflictOpts.ERROR):
```
ebrahimebrahim, line 195: this is a good time to add a docstring to this function explaining for example what is meant by registration surface and body model

src/openlifu/db/database.py
```diff
@@ -547,6 +571,22 @@ def get_photoscan_info(self, subject_id, session_id, photoscan_id):
...
+    def get_transducer_info(self, transducer_id):
+        transducer_metadata_filepath = self.get_transducer_filename(transducer_id)
+        with open(transducer_metadata_filepath) as f:
+            transducer = json.load(f)
+            transducer_dict = {"id": transducer["id"],\
+                    "name": transducer["name"],\
+                    "elements": transducer["elements"], \
+                    "frequency": transducer["frequency"], \
+                    "units": transducer["units"], \
+                    "attrs": transducer["attrs"]}
```
ebrahimebrahim, line 583: What's the reasoning behind including all these other things here besides the absolute filepaths? Since these could be obtained by reading the `Transducer` object from the database, maybe there is no reason to also include them here. I'm referring to things like elements, frequency, units, attrs. LMK what you think I might not be fully aware of a future plan for this

---

If the main purpose of `get_transducer_info` is to get the absolute filepaths of the affiliated data files like the body models, then maybe a better name for this function is `get_transducer_absolute_filepaths` and we can focus it on just doing that.

This function `get_transducer_info`  could also use a docstring to help explain when/why one would want to use this versus `load_transducer`

src/openlifu/db/database.py
```diff
@@ -208,6 +208,30 @@ def write_transducer(self, transducer, on_conflict: OnConflictOpts=OnConflictOpt
...
 
         transducer_filename = self.get_transducer_filename(transducer_id)
+        transducer_parent_dir = transducer_filename.parent
+        transducer_parent_dir.mkdir(exist_ok = True)
+
+        # Copy the transducer data files to database.
+        # If tranducer data files not provided, check that any files previously
+        # associated with the transducer object exist.
+        if registration_surface_model_filepath:
+            if not Path(registration_surface_model_filepath).exists():
```
ebrahimebrahim, line 218: setting `registration_surface_model_filepath = Path(registration_surface_model_filepath)` earlier on can help avoid having to repeatedly call the constructor `Path()` below. similar for `transducer_body_model_filepath`

ebrahimebrahim (changes requested):
Very nice and good unit tests, just some simple changes requested

src/openlifu/db/database.py
sadhana-r: I had included all of the attributes to be consistent with other similar functions like `get_photoscan_info` or `get_volume_info`. But I agree that the main use case is to get the absolute filepaths and returning all of the data like elements, frequency is overkill and likely not going to be used in this way. I'll rename the function so it doesn't get confused with the other `get_info` functions and just return the filepaths.

db_dvc.dvc
sadhana-r: Yep seems like it. I ran `dvc push` again. Let me know if you still run into an error.

src/openlifu/db/database.py
```diff
@@ -193,6 +193,10 @@ def write_subject(self, subject, on_conflict=OnConflictOpts.ERROR):
         self.logger.info(f"Added subject with ID {subject_id} to the database.")
 
     def write_transducer(self, transducer, registration_surface_model_filepath: Optional[str] = None, transducer_body_model_filepath: Optional[str] = None, on_conflict: OnConflictOpts=OnConflictOpts.ERROR):
+        """ Writes a transducer object to database and copies the affiliated transducer data files i.e. transdcer body and
+        registration surface model files to the database.These model files are optional data attributes that are only required if running
+        transducer tracking. When a transducer that is already present in the database is being re-written,
+        the associated transducer model files do not need to be provided """
```
ebrahimebrahim, line 199: It would be good to explain in this docstring what is meant by registration surface and body model, because it's really not clear just from the variable name what's the intention.

I suggest a docstring style where you lay out all the args and return value one by one like we have been doign with some of the other functions

db_dvc.dvc
ebrahimebrahim: all good now :+1: 

(but a merge of the db_dvc data might be needed by the time we are ready to merge this pr)

src/openlifu/db/database.py
```diff
@@ -547,6 +577,23 @@ def get_photoscan_info(self, subject_id, session_id, photoscan_id):
                 photoscan_dict["mtl_abspath"] = Path(photoscan_metadata_filepath).parent/photoscan["mtl_filename"]
             return photoscan_dict
 
+    def get_transducer_absolute_filepaths(self, transducer_id):
+        """ Returns the absolute filepaths for the data files i.e.
+        transducer body and registration surface model files affiliated with the transducer"""
```
ebrahimebrahim, line 582: I think this docstring needs a bit more to help explain the _intention_: why should one want to use this instead of `load_transducer`? How is this related to `load_transducer`?

Same for the `load_transducer` docstring -- it should describe its own limitations and its relationship to `get_transducer_absolute_filepaths`

(These kinds of explanations are important for usability I think because at some point (soon I hope) we will set up a readthedocs which will generate doc pages out of these docstrings)

Finally, this docstring here should describe clearly what is returned (a dict with such and such keys)

(And add type annotations also while we are at it)

ebrahimebrahim (changes requested):
More docstring nitpicks :grinning:

ebrahimebrahim:
Ah one  other small thing -- that last commit message title is very long, there should be a 50-72 char limit on the title of a commit msg

db_dvc.dvc
sadhana-r: I merged db_dvc now as well.

ebrahimebrahim (commented):
The docstrings look great!

The dvc data I think hasn't been properly rebased to the _current_ main

sadhana-r:
> The dvc data I think hasn't been properly rebased to the _current_ main

Whoops, not sure where I went wrong when I was doing it before. Fixed it now!

Issue #167, opened 2024-11-20
Continue to support python 3.9 


ebrahimebrahim:
k-wave-python [dropped python 3.9 support](https://github.com/waltsims/k-wave-python/pull/505) in [the latest release](https://github.com/waltsims/k-wave-python/releases/tag/v0.3.6)

We may have to pin the k-wave-python version or maintain a parallel fork, because OpenLIFU-python needs to support python3.9 in order to be used in 3D Slicer.

arhowe00:
We should pin the version of kwave to 0.3.4. k-wave-python 0.3.5 is automatically installed as a dependency, which introduces the following bug:

```
Traceback (most recent call last):
  File "./SlicerOpenLIFU-Debug/lib/Slicer-5.7/qt-scripted-modules/OpenLIFUSonicationPlanner.py", line 300, in onComputeSolutionClicked
    self.logic.computeSolution(activeData["Volume"], activeData["Target"],
  File "./SlicerOpenLIFU-Debug/lib/Slicer-5.7/qt-scripted-modules/OpenLIFUSonicationPlanner.py", line 519, in computeSolution
    solution_openlifu, pnp_aggregated, intensity_aggregated, analysis_openlifu = compute_solution_openlifu(
  File "./SlicerOpenLIFU-Debug/lib/Slicer-5.7/qt-scripted-modules/OpenLIFUSonicationPlanner.py", line 479, in compute_solution_openlifu
    solution, simulation_result_aggregated, scaled_solution_analysis = protocol.calc_solution(
  File "./Slicer-Debug/python-install/lib/python3.9/site-packages/openlifu/plan/protocol.py", line 233, in calc_solution
    simulation_output_xarray, _ = run_simulation(
  File "./Slicer-Debug/python-install/lib/python3.9/site-packages/openlifu/sim/kwave_if.py", line 115, in run_simulation
    p_max = xa.DataArray(output['p_max'].reshape(sz, order='F'),
KeyError: 'p_max'
```

The aforementioned bug was already fixed in the following commit released in k-wave-python:
```
commit 0ab1e87a4bebbe5042c476c67bb0a8f08f936d3c
Author: brubbel <5667690+brubbel@users.noreply.github.com>
Date:   Tue Nov 19 20:12:48 2024 +0100

    bugfix simulation command options (#509)

    fixes https://github.com/waltsims/k-wave-python/issues/376
```

but this was released with a version that also does not Python 3.9.

PR #193, opened 2025-01-21
Pin k-wave-python 0.3.4 to avoid a bug in 0.3.5

ebrahimebrahim:
When using python 3.9, which we need to be able to do as explained in #167, the latest version of k-wave-python that can be installed is 0.3.5, and this has a bug that has been affecting our simulation runs. The fix to the bug takes place in a later version of k-wave-python that has dropped python 3.9 support.

Ideally we should unit test the simulation running code in order to have caught this. ~~But punting on that for now.~~ Never mind I did end up adding a test!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,6 +55,10 @@ jobs:
           python-version: ${{ matrix.python-version }}
           allow-prereleases: true
 
+      - name: Install system dependencies
+        if: runner.os == 'macOS'
+        run: brew install hdf5 fftw zlib libomp
+
       - name: Install package
         run: python -m pip install .[test]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,7 +42,7 @@ dependencies = [
   "nibabel",
   "sphinx",
   "ipykernel",
-  "k-wave-python>=0.3.4",
+  "k-wave-python==0.3.4",
   "nvidia-ml-py",
   "OpenEXR"
 ]
--- a/tests/test_sim.py
+++ b/tests/test_sim.py
@@ -1,0 +1,45 @@
+import xarray
+
+import openlifu
+
+
+def test_run_simulation_runs():
+    """Test that run_simulation can run and outputs something of the correct type."""
+
+    transducer = openlifu.Transducer.gen_matrix_array(nx=2, ny=2, pitch=4, kerf=.5, units="mm", impulse_response=1e5)
+    dt = 2e-7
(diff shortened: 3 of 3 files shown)


ebrahimebrahim:
That's interesting, the test revealed an issue on mac -- that's the same thing you had run into earlier right @arhowe00 ?

And the solution was to install hdf5, fftw, zlib, and libomp at the system level?

Hmm I wonder if there is an extra setup step needed in our https://github.com/OpenwaterHealth/OpenLIFU-python/blob/main/.github/workflows/ci.yml

k-wave-python appear to run their unit tests on mac
https://github.com/waltsims/k-wave-python/blob/master/.github/workflows/pytest.yml#L76
but I don't see any place where install hdf5, fftw, etc.
Need to look more into this

arhowe00:
> That's interesting, the test revealed an issue on mac -- that's the same thing you had run into earlier right @arhowe00 ?
> 
> And the solution was to install hdf5, fftw, zlib, and libomp at the system level?

That's correct. 
 
> Hmm I wonder if there is an extra setup step needed in our https://github.com/OpenwaterHealth/OpenLIFU-python/blob/main/.github/workflows/ci.yml
> 
> k-wave-python appear to run their unit tests on mac https://github.com/waltsims/k-wave-python/blob/master/.github/workflows/pytest.yml#L76 but I don't see any place where install hdf5, fftw, etc. Need to look more into this

`k-wave-python` should ideally bundle the dependencies when doing `pip install k-wave-python`.  I am not sure which step exposes the required libraries for `run_simulation` to the linker, but it could be the line `pip install '.[test]'` which installs the dependencies that allow their tests to pass.

We could modify the CI pipeline to install the necessary libraries on macOS,

```
- name: Install system dependencies
  run: |
    brew install hdf5 fftw zlib libomp
```

but this is a fragile solution.

ebrahimebrahim:
Thanks for figuring this one out!!

So
- I will add your fix to the CI here.
- I think k-wave-python should be documenting the required system dependencies on mac when it isn't obvious like this -- so I think you should open an issue @arhowe00 on the k-wave-python repo asking them if these dependencies really are needed on mac (i.e. we haven't missed something here) and suggesting adding that to their docs if so.

ebrahimebrahim:
@arhowe00 It is ready, please review

arhowe00 (approved):

arhowe00:
I opened up an issue here on [k-wave-python#549](https://github.com/waltsims/k-wave-python/issues/new?template=bug_report.md)

PR #197, opened 2025-01-28
Add to class Database and test delete_protocol

### LINKED ISSUE #196: Add the ability to delete protocols to Database
The current implementation of `Database` in `src/openlifu/db/database.py` does not support deleting protocols objects from the JSON database.  Currently, there is support for the following functions: 

- `write_protocol(self, protocol: Protocol, on_conflict: OnConflictOpts = OnConflictOpts.ERROR)`
- `write_run(self, run:Run, session:Session = None, protocol:Protocol = None, on_conflict=OnConflictOpts.ERROR)`
- `get_protocol_ids(self)`
- `load_protocol(self, protocol_id)`
- `load_all_protocols(self)`
- `load_protocol_snapshot(self, subject_id, session_id, run_id)`
- `get_protocols_filename(self)`
- `get_protocol_filename(self, protocol_id)`
- `write_protocol_ids(self, protocol_ids)`

arhowe00:
- Closes #196

### DIFF
--- a/src/openlifu/db/database.py
+++ b/src/openlifu/db/database.py
@@ -71,6 +71,29 @@ class Database:
 
         self.logger.info(f"Added Sonication Protocol with ID {protocol_id} to the database.")
 
+    def delete_protocol(self, protocol_id: str, on_conflict: OnConflictOpts = OnConflictOpts.ERROR):
+        # Check if the sonication protocol ID already exists in the database
+        protocol_ids = self.get_protocol_ids()
+
+        if protocol_id not in protocol_ids:
+            if on_conflict == OnConflictOpts.ERROR:
+                raise ValueError(f"Protocol ID {protocol_id} does not exist in the database.")
+            elif on_conflict == OnConflictOpts.SKIP:
+                self.logger.info(f"Cannot delete protocol ID {protocol_id} as it does not exist in the database.")
+                return
+            else:
+                raise ValueError("Invalid 'on_conflict' option.")
+
+        # Delete the directory of the protocol
+        protocol_dir = self.get_protocol_dir(protocol_id)
+        if Path.is_dir(protocol_dir):
+            shutil.rmtree(protocol_dir)
+
+        if protocol_id in protocol_ids:
+            protocol_ids.remove(protocol_id)
+            self.write_protocol_ids(protocol_ids)
+
+        self.logger.info(f"Removed Sonication Protocol with ID {protocol_id} from the database.")
 
     def write_session(self, subject:Subject, session:Session, on_conflict=OnConflictOpts.ERROR):
         # Generate session ID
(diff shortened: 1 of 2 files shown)


src/openlifu/db/database.py
```diff
@@ -71,6 +71,28 @@ def write_protocol(self, protocol: Protocol, on_conflict: OnConflictOpts = OnCon
...
+        protocol_ids = self.get_protocol_ids()
+
+        if protocol_id not in protocol_ids:
+            if on_conflict == OnConflictOpts.ERROR:
+                raise ValueError(f"Protocol ID {protocol_id} does not exist in the database.")
+            elif on_conflict == OnConflictOpts.SKIP:
+                self.logger.info(f"Cannot delete protocol ID {protocol_id} as it does not exist in the database.")
+                return
+            else:
+                raise ValueError("Invalid 'on_conflict' option. Use 'error', 'overwrite', or 'skip'.")
```
ebrahimebrahim, line 85: can you add test coverage of these cases?

ebrahimebrahim, line 85:
```py
raise ValueError("Invalid 'on_conflict' option. Use 'error', 'overwrite', or 'skip'.")
```

also "overwrite" wouldn't make sense here

in fact this whole message doesn't make sense-- it is there in other places as a holder from back when we had strings instead of the `OnConflictOpts` enum. better to just say `"Invalid 'on_conflict' option.` and leave it at that

src/openlifu/db/database.py
```diff
@@ -71,6 +71,28 @@ def write_protocol(self, protocol: Protocol, on_conflict: OnConflictOpts = OnCon
...
+
+        # Delete the sonication protocol JSON file
+        protocol_filename = self.get_protocol_filename(protocol_id)
+        Path.unlink(protocol_filename)
```
ebrahimebrahim, line 89: this deletes the file but is there a dangling folder left behind? (I didn't test it I am just thinking there may be a dangling empty folder left behind)

ebrahimebrahim (changes requested):

src/openlifu/db/database.py
arhowe00: It appears so -- which I will remove. Not sure why protocols are enclosed within folders in the first place but I will assume that's what we wanted.

ebrahimebrahim: > Not sure why protocols are enclosed within folders in the first place

It was like that when we got here :shrug:  

Maybe the idea is to leave open the possibility of having data file affiliated with protocols

src/openlifu/db/database.py
arhowe00: I added coverage of these cases/

ebrahimebrahim (approved):
Looks good!
The commit message

> Add and test delete_protocol to class Database

is grammatically weird... can you make it sound a bit more natural? Maybe just

> Add Protocol deletion to Database

or similar

Then feel free to merge!

arhowe00:
> ...
> > Add and test delete_protocol to class Database
> 
> is grammatically weird... can you make it sound a bit more natural? Maybe just
> ...

I agree. I guess you know this one wasn't written by AI!

PR #202, opened 2025-02-07
Added docstrings to Protocol.py

### LINKED ISSUE #201: Add docstrings to Protocol attributes
Add docstrings under the attributes of `Protocol`

https://github.com/OpenwaterHealth/OpenLIFU-python/blob/af3fc5741ca03d4eb87f9f10f614f5f2279efdef/src/openlifu/plan/protocol.py#L30-L42

An example is how we have docstrings under the attributes of `Solution`

https://github.com/OpenwaterHealth/OpenLIFU-python/blob/af3fc5741ca03d4eb87f9f10f614f5f2279efdef/src/openlifu/plan/solution.py#L33-L53

These docstrings will end up listed on the [online doc page for protocol](https://openlifu.readthedocs.io/en/latest/_autosummary/openlifu.plan.protocol.Protocol.html#openlifu.plan.protocol.Protocol), and we can also use them as tooltip text on the respective fields in the SlicerOpenLIFU protocol configuration module

Where certain physical units are expected, this should also be expressed in the docstring

peterhollender:
Closes #201

### DIFF
--- a/src/openlifu/plan/protocol.py
+++ b/src/openlifu/plan/protocol.py
@@ -28,18 +28,43 @@ OnPulseMismatchAction = Enum("OnPulseMismatchAction", ["ERROR", "ROUND", "ROUNDU
 @dataclass
 class Protocol:
     id: str = "protocol"
+    """ The unique identifier of the protocol """
+
     name: str = "Protocol"
+    """ The name of the protocol """
+
     description: str = ""
+    """ A more detailed description of the protocol """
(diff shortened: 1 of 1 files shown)


ebrahimebrahim:
Can you add the issue number (#201) to the commit message (header or body) and `git commit --amend` and force push?

ebrahimebrahim (approved):

PR #211, opened 2025-02-12
Add sonication control API

ebrahimebrahim:
This copies over sonication control API files that George pointed out by email, taken from the `p/sonication_control_refactor` branch which was at commit 0478491ee6be2fda76b7e7d700a7a205244293f4 at the time of  copying.

@georgevigelette and @peterhollender I will be putting some questions for you here before I can fix this up for final review

@sadhana-r and I will be ensuring that this provides the needed API for SlicerOpenLIFU to use

---

Additional slightly paraphrased notes from George:

[DemoApplication.py](https://github.com/OpenwaterHealth/OpenLIFU-python/blob/0478491ee6be2fda76b7e7d700a7a205244293f4/notebooks/DemoApplication.py) s the old minimalized pyQT app for testing that has been used for some time and works with this library

(there are possibly additional useful scripts to reference in [the notebooks directory of p/sonication_control_refactor](https://github.com/OpenwaterHealth/OpenLIFU-python/tree/0478491ee6be2fda76b7e7d700a7a205244293f4/notebooks).)

You should only really be interfacing with LIFUInterface.py where all the lower level configurations will be done for you from a solution.  If you look at the pyQT app you will see that it is setting the registers by accessing the transmitter directly which you shouldn't need to do.

You will need to do something like the following to initialize the interface (setting `test_mode` to True used to work, haven't tested it but it would just spit back that the hardware is there so you can use the interface):

```py
from openlifu.io.LIFUInterface import LIFUInterface

def init_lifu(self):
        self.interface = LIFUInterface(test_mode=False)
        tx_connected, hv_connected = self.interface.is_device_connected()
        if tx_connected and hv_connected:
            print("LIFU Device Fully connected.")
        else:
            print(f"LIFU Device NOT Fully Connected. TX: {tx_connected}, HV: {hv_connected}")
```

You will use it like below, providing the solution and starting sonication:
```py
self.interface.set_solution(solution= None)     # Set solution
self.interface.get_solution()                   # Get solution
self.interface.start_sonication()               # Start sonication
self.interface.stop_sonication()                # Stop sonication
```

There is a method for monitoring and some others that can be used.


src/openlifu/io/__init__.py
```diff
@@ -1,19 +1,16 @@
...
-    "Tx7332Registers",
-    "TxModule",
-    "TxArray",
-    "print_regs",
-    "swap_byte_order",
+    # "PulseProfile",
+    # "DelayProfile",
+    # "Tx7332Registers",
+    # "print_regs",
+    # "swap_byte_order",
```
ebrahimebrahim, line 14: I will be removing these comments

src/openlifu/io/__init__.py
```diff
@@ -1,19 +1,16 @@
...
-    print_regs,
-    swap_byte_order,
-)
+from openlifu.io.config import *  # noqa: F403
+from openlifu.io.i2c_data_packet import *  # noqa: F403
+from openlifu.io.i2c_status_packet import *  # noqa: F403
+from openlifu.io.LIFUInterface import LIFUInterface
+from openlifu.io.LIFUUart import *  # noqa: F403
+from openlifu.io.tx7332_if import *  # noqa: F403
+from openlifu.io.utils import *  # noqa: F403
```
ebrahimebrahim, line 7: I will be fixing these imports and removing the noqa

src/openlifu/io/ustx.py
```diff
@@ -403,7 +404,7 @@ def get_delay_control_registers(self, profile: Optional[int]=None) -> Dict[int,i
         delay_profile = self.get_delay_profile(profile)
         apod_register = 0
         for i, apod in enumerate(delay_profile.apodizations):
-            apod_register = set_register_value(apod_register, 1-apod, lsb=i, width=1)
+            apod_register = set_register_value(apod_register, 1-apod, lsb=APODIZATION_CHANNEL_ORDER.index(i+1), width=1)
```
ebrahimebrahim, line 407: @georgevigelette @peterhollender Please check this change introduced into `main` from `p/sonication_control_refactor`. Do we want this?

src/openlifu/io/ustx.py
```diff
@@ -481,7 +482,7 @@ def get_pulse_data_registers(self, profile: Optional[int]=None, pack: bool=False
         levels = pattern['levels']
         lengths = pattern['lengths']
         nperiods = len(levels)
-        level_lut = {-1: 0b01, 0: 0b00, 1: 0b10}
+        level_lut = {-1: 0b01, 0: 0b11, 1: 0b10}  # Map levels to register values 0b11 drive to ground 0b00 high impedance
```
ebrahimebrahim, line 485: @georgevigelette @peterhollender Please check this change introduced into `main` from `p/sonication_control_refactor`. Do we want this?

src/openlifu/io/i2c_data_packet.py
```diff
@@ -0,0 +1,125 @@
...
+            raise ValueError("CRC validation failed.")
+        
+        # Set CRC to the received CRC
+        self.crc = received_crc
+
+        return self
+
+
+    def print_packet(self):
+        print("I2C Data Packet")
```
ebrahimebrahim, line 68: I will replace this and other print statements by logging.

src/openlifu/io/LIFUUart.py
```diff
@@ -0,0 +1,345 @@
...
+    OW_CMD_NOP,
+    OW_END_BYTE,
+    OW_START_BYTE,
+)
+from openlifu.io.LIFUSignal import LIFUSignal
+from openlifu.io.uartpacket import UartPacket
+from openlifu.io.utils import util_crc16
+
+# Set up logging
+logging.basicConfig(level=logging.DEBUG)
```
ebrahimebrahim, line 21: Setting the root logger to debug level at the top of this file is not appropriate as this will affect logging globally just when LIFUUart gets imported. I will adjust this.

src/openlifu/io/tx7332_if.py
```diff
@@ -0,0 +1,372 @@
+import logging
+import struct
+from typing import List
+
+logging.basicConfig(level=logging.DEBUG)
```
ebrahimebrahim, line 5: Setting the root logger to debug level at the top of this file is not appropriate as this will affect logging globally just when tx7332_if.py gets imported. I will adjust this.

src/openlifu/io/LIFUUart.py
```diff
@@ -0,0 +1,345 @@
...
+        self.pid = pid
+        self.port = None
+        self.baudrate = baudrate
+        self.timeout = timeout
+        self.align = align
+        self.packet_count = 0
+        self.serial = None
+        self.running = False
+        self.asyncMode = False
+        self.loop = asyncio.get_event_loop()
```
ebrahimebrahim, line 47: @georgevigelette Currently if I try to instantiate a `LIFUInterface` I get an error at this line:

```txt
RuntimeError: There is no current event loop in thread 'MainThread'.
```

This happens whether I set `test_mode=True` or not

georgevigelette:
That loop isn't needed and I just pushed a change that removed the
requirements.

ebrahimebrahim:
> That loop isn't needed and I just pushed a change that removed the requirements.

@georgevigelette A lot [changed](https://github.com/OpenwaterHealth/OpenLIFU-python/compare/0478491ee6be2fda76b7e7d700a7a205244293f4..9773a7a5eccddbad7038496341f67c811f331e4a) in those changes besides just the event loop thing:
- `LIFUTXDevice.py`
- `LIFUUart.py`
- `config.py`

It's not clear to me what to port over from those changes.

Also, will you continue adding things to `p/sonication_control_refactor` or will you eventually move your development to `main` after this PR is merged? If you will keep working on `p/sonication_control_refactor` then we need to go back and work on merging _that_ branch into main, right? @peterhollender can you help explain what's going on with the branching here?

ebrahimebrahim:
Close this. Move to a PR that merges p/sonication_control_refactor into main.
Delete files that we want to remove (the ones not included here)

We want everything that has happened in this PR plus the pulse fix (Pa-> VOlts) that happened in that branch.

PR #212, opened 2025-02-13
Incorporate virtual fit results into Session (#179)

### LINKED ISSUE #179: Add virtual fit transform to session definition
The transducer transform resulting from virtual fit should be saved to sessions as a special thing, not just as the “current transducer transform” alongside a virtual fit approval flag. One reason for this is that later in the transducer tracking workflow we want to be able to show a comparison and clear visual deviation between the virtual fit transform and the transducer tracking transform. 

So the goal here is to replace the current virtual fit approval flag/target in `Session` by a richer data structure consisting of
- Virtual fit approval state
- Target ID of the virtual fit
- _Transducer transform_ resulting from the virtual fit

UPDATE: Actually it's not  one transform but the entire collection of maybe the top N transforms from the virtual fit search, sorted so that the first one is the best. The reason for this is that we want to use all the trasforms as possible initializations for transducer surface registration.

ebrahimebrahim:
Closes #179 

## `Session.virtual_fit_results`

Virtual fit results take the form of a dictionary mapping target IDs
to pairs `(approval, transforms)`, where:

- `approval` is a boolean indicating whether the virtual fit for that target has been approved, and
- `transforms` is a list of transducer transforms resulting from the virtual fit for that target.

The idea is that the list of transforms would be ordered from best to worst, and should of course
contain at least one transform.

## `standardize_database`

This is a utility script for developers to read in and write back out the dvc database. It is useful for standardizing the format of the example dvc data, and also for checking that the database mostly still works.

I have been using this on my own to help catch little discrepancies that pop up between the example data and the openlifu library -- I figured it's probably good to version control it here under a `scripts/` folder since it needs to evolve alongside the library.

To use this script, install openlifu to a python environment and then run the script providing the database folder as an argument:

```
python standardize_database.py db_dvc/
```

## Parallel SlicerOpenLIFU PR

(todo)

## Remaining to-do

- [x] Update DVC data
- [x] Start work on a parallel PR on SlicerOpenLIFU that makes the virtual fitting placeholder be compatible with this breaking change
  - It is here: https://github.com/OpenwaterHealth/SlicerOpenLIFU/pull/179

### DIFF
--- a/scripts/standardize_database.py
+++ b/scripts/standardize_database.py
@@ -79,3 +79,9 @@ for subject_id in db.get_subject_ids():
         run_ids = db.get_run_ids(subject_id, session_id)
         db.write_run_ids(subject_id, session_id, run_ids)
         # (Runs are read only at the moment so it's just the runs.json and no individual runs to standardize)
+
+db.write_user_ids(db.get_user_ids())
+for user_id in db.get_user_ids():
+    user = db.load_user(user_id)
+    assert user_id == user.id
+    db.write_user(user, on_conflict=OnConflictOpts.OVERWRITE)
--- a/scripts/standardize_database.py
+++ b/scripts/standardize_database.py
@@ -1,0 +1,81 @@
+"""This is a utility script for developers to read in and write back out the dvc database.
+It is useful for standardizing the format of the example dvc data, and also for checking that the database
+mostly still works.
+
+To use this script, install openlifu to a python environment and then run the script providing the database folder as an argument:
+
+```
+python standardize_database.py db_dvc/
+```
+
+A couple of known issues to watch out for:
+- The date_modified of Sessions gets updated, as it should, when this is run. But we don't care about that change.
+- The netCDF simulation output files (.nc files) are modified for some reason each time they are written out. It's probably a similar
+  thing going on with some kind of timestamp being embedded in the file.
+"""
+
+import pathlib
+import shutil
+import sys
+import tempfile
+
+from openlifu.db import Database
+from openlifu.db.database import OnConflictOpts
+
+if len(sys.argv) != 2:
+    raise RuntimeError("Provide exactly one argument: the path to the database folder.")
+db = Database(sys.argv[1])
+
+db.write_protocol_ids(db.get_protocol_ids())
+for protocol_id in db.get_protocol_ids():
+    protocol = db.load_protocol(protocol_id)
+    assert protocol_id == protocol.id
+    db.write_protocol(protocol, on_conflict=OnConflictOpts.OVERWRITE)
+
+db.write_transducer_ids(db.get_transducer_ids())
+for transducer_id in db.get_transducer_ids():
+    transducer = db.load_transducer(transducer_id)
+    assert transducer_id == transducer.id
+    db.write_transducer(transducer, on_conflict=OnConflictOpts.OVERWRITE)
+
+db.write_subject_ids(db.get_subject_ids())
+for subject_id in db.get_subject_ids():
+    subject = db.load_subject(subject_id)
+    assert subject_id == subject.id
+    db.write_subject(subject, on_conflict=OnConflictOpts.OVERWRITE)
+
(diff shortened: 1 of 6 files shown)


ebrahimebrahim:
@arhowe00 This PR is coming in parallel to https://github.com/OpenwaterHealth/SlicerOpenLIFU/pull/179 which probably @sadhana-r will end up reviewing. But this one should be merged first and is completed, so you can start reviewing it. Besides the new virtual fit result format, I lumped in a script I've been using from time to time to make sure we haven't broken the database. LMK if you have any questions or if you want to have a quick meeting to clarify what's going on with virtual fitting and all that. TY!

scripts/standardize_database.py
```diff
@@ -0,0 +1,81 @@
...
+        db.write_solution_ids(session, solution_ids)
+        for solution_id in solution_ids:
+            solution = db.load_solution(session, solution_id)
+            assert solution.id == solution_id
+            assert solution.simulation_result['p_min'].shape[0] == solution.num_foci()
+            db.write_solution(session, solution, on_conflict=OnConflictOpts.OVERWRITE)
+
+        run_ids = db.get_run_ids(subject_id, session_id)
+        db.write_run_ids(subject_id, session_id, run_ids)
+        # (Runs are read only at the moment so it's just the runs.json and no individual runs to standardize)
```
arhowe00, line 81:
```suggestion
        # (Runs are read only at the moment so it's just the runs.json and no individual runs to standardize)
        
db.write_user_ids(db.get_user_ids())
for user_id in db.get_user_ids():
    user = db.load_user(user_id)
    assert user_id == user.id
    db.write_user(user, on_conflict=OnConflictOpts.OVERWRITE)
```
Since the users table was merged into main, you might want to add a bit for that. The way it works is relatively simple so I think this suggestion should work.

src/openlifu/db/session.py
```diff
@@ -128,6 +137,12 @@ def from_dict(d:Dict):
             d['targets'] = [Point.from_dict(d['targets'])]
         elif isinstance(d['targets'], Point):
             d['targets'] = [d['targets']]
+        if 'virtual_fit_results' in d:
```
arhowe00, line 140: I know you did not touch the code in the function signature in this PR, but I noticed that some of the docs are outdated, such as:

```
 :param db: Database object
```

While it never makes sense to overhaul the whole file if you just touch one part of it, it may be worth leaving unrelated fixes in your trails... I'm going to try making a commit to this PR that addresses this thing I noticed, since it would be more efficient for me to do that right now

arhowe00:
Great job, I had just a few minor comments and questions. Also suggested adding checking the users table in your script.

scripts/standardize_database.py
ebrahimebrahim: thanks!

src/openlifu/db/session.py
ebrahimebrahim: Thanks! I included the fix. As we discussed it's up to us to decide what's the etiquette on that sort of thing... I didn't feel it was too intrusive or anything in this case. You can also feel free to make tons of small PRs for trivial things. It's not at all bad to that and I don't mind reviewing lots of tiny PRs. And also as we discussed not every PR necessarily requires opening a new issue since there are often existing issues to reference (but lmk if there isn't one for something)

ebrahimebrahim:
(The rebase required merging the dvc data changes in main -- it was just the stuff you added in #214 -- should be good now)

arhowe00 (approved):

ebrahimebrahim:
Note to self: Do not merge yet please! I want to finish off https://github.com/OpenwaterHealth/SlicerOpenLIFU/pull/179 first since this is a breaking change for SlicerOpenLIFU, and I don't want to cause trouble for other work you guys are doing on SlicerOpenLIFU

ebrahimebrahim:
@arhowe00 I found a small issue and just pushed a fix, the commit message should explain it. Letting you know since you had already approved. Thanks!

Issue #205, opened 2025-02-07
Add sonication control API


ebrahimebrahim:
Add an API to communicate with hardware for sonication control

Should be able to

- pass in a solution
- control the hardware (send start/stop signals, etc.)
- monitor hardware (get progress, etc.)

Issue #75, opened 2024-07-08
Sonication Control/Control Sonication (SOFTREQ-72)


peterhollender:
The software must be able to initiate, pause, and abort sonication

Notes:
nan

Updated on 2024-07-08 15:29:34

PR #213, opened 2025-02-13
The great merge of p/sonication_control_refactor into main

ebrahimebrahim:
When this is ready we will do a squash-merge into a single commit, referencing #205 and #75 in the commit message.


README.rst
```diff
@@ -87,6 +87,10 @@ To commit updates to the sample database:
    git push
    dvc push #Requires access to remote storage
 
+Set Python Path
+.. code:: sh
+   set PYTHONPATH=%cd%\src
+
```
ebrahimebrahim, line 93: This doesn't belong here I think, it seems too specific to a particular development environment and in any case would belong in the installation section above or in the contributing guide. @peterhollender May I remove it?

src/openlifu/io/__init__.py
```diff
@@ -1,19 +1,14 @@
...
-    "Tx7332Registers",
-    "TxModule",
-    "TxArray",
-    "print_regs",
-    "swap_byte_order",
+    # "PulseProfile",
+    # "DelayProfile",
+    # "Tx7332Registers",
+    # "print_regs",
+    # "swap_byte_order",
```
ebrahimebrahim, line 12: I will remove these comments

src/openlifu/io/__init__.py
```diff
@@ -1,19 +1,14 @@
...
-    TxArray,
-    TxModule,
-    print_regs,
-    swap_byte_order,
-)
+from openlifu.io.config import *  # noqa: F403
+from openlifu.io.LIFUInterface import LIFUInterface
+from openlifu.io.LIFUUart import *  # noqa: F403
+from openlifu.io.tx7332_if import *  # noqa: F403
+from openlifu.io.utils import *  # noqa: F403
```
ebrahimebrahim, line 5: I will remove noqa and fix the imports

src/openlifu/io/tx7332_if.py
```diff
@@ -0,0 +1,372 @@
+import logging
+import struct
+from typing import List
+
+logging.basicConfig(level=logging.DEBUG)
```
ebrahimebrahim, line 5: Setting the root logger to debug level at the top of this file is not appropriate as this will affect logging globally just when the file gets imported. I will adjust this.

src/openlifu/io/LIFUUart.py
```diff
@@ -0,0 +1,337 @@
...
+    OW_END_BYTE,
+    OW_ERROR,
+    OW_START_BYTE,
+)
+from openlifu.io.LIFUSignal import LIFUSignal
+from openlifu.io.uartpacket import UartPacket
+from openlifu.io.utils import util_crc16
+
+# Set up logging
+logging.basicConfig(level=logging.DEBUG)
```
ebrahimebrahim, line 21: Setting the root logger to debug level at the top of this file is not appropriate as this will affect logging globally just when the file gets imported. I will adjust this.

pyproject.toml
```diff
@@ -42,6 +42,10 @@ dependencies = [
   "nibabel",
   "sphinx",
   "ipykernel",
+  "pyserial",
+  "PyQt5",
+  "crcmod",
+  "qasync",
```
ebrahimebrahim, line 48: Do we need PyQt5 and qasync in openlifu? @peterhollender I will remove these -- I imagine were for the demo application right? We can introduce a separate optional dependency group `[demoapp]` if we want to make in convenient to set up an environment in which the demo qt application works. What do you think @peterhollender ?

README.rst
ebrahimebrahim: Actually I just looked at this more carefully, it looks like this is to avoid having to _install_ openlifu. But we should be telling our users to install it, not to do this.

For a developers the same thing can be achieved as this, but better, if they do an _editable_ install (`pip install -e .` as shown above in the readme). So I will remove this.

src/openlifu/io/LIFUInterface.py
```diff
@@ -0,0 +1,160 @@
...
+        """
+        Load a solution to the device.
+
+        Args:
+            solution (Solution): The solution to load.
+        """
+        try:
+            logger.info("Loading solution: %s", solution.name)
+            # Convert solution data and send to the device
+            self.devicecontroller.set_solution(solution)
```
ebrahimebrahim, line 88: @georgevigelette @peterhollender  When I try to use `LIFUInterface.set_solution`, I get an error here because `devicecontroller` is not defined

pyproject.toml
peterhollender: @georgevigelette is the `async` stuff _purely_ in the demo app? Or are there parts of this code that use it, but just aren't _useful_ outside of the app context? If the slicer module does use the async functionality, will all those `async` calls be on the slicer side, or should this remain a dependency?

ebrahimebrahim: (note that the general async stuff using `asyncio` comes in the python stdlib -- here I am removing [qasync](https://pypi.org/project/qasync/) which is qt specific)

src/openlifu/io/utils.py
```diff
@@ -0,0 +1,71 @@
...
+# ruff: noqa: T201
+def format_and_print_hex(data):
+    """
+    Format the received data as hex and print it.
+    """
+    if data:
+        hex_output = ' '.join([f'{byte:02X}' for byte in data])
+        print("Received Data (Hex):", hex_output)
+    else:
+        print("Received Data (Hex): EMPTY")
```
ebrahimebrahim, line 56: I opted to ignore the print statements here in the CI rather then using a logger, since this function is clearly a debug utility that is explicitly supposed to print

src/openlifu/io/LIFUUart.py
```diff
@@ -0,0 +1,340 @@
...
+            if self.asyncMode:
+                # Start the reading thread
+                self.read_thread = threading.Thread(target=self._read_data)
+                self.read_thread.daemon = True
+                self.read_thread.start()
+        except Exception as e:
+            log.error(f"Failed to connect to {self.port}: {e}")
+            self.running = False
+            self.port = None
+            raise e
```
ebrahimebrahim, line 83: Note that in several places, such as as here, I added this to re-raise the exception. Catching and trashing general exceptions is not the responsibility of this code -- if the application _using_ this code wants to not crash when it hits the exception then it can catch it and handle it.

If there is a more specific type of exception besides the broadest "`Exception`" type that we deliberately want to throw out for good reason here, then we can specify it instead of catching all Exceptions

LMK if the decision to throw out the exception was deliberate and my change here breaks something -- I am not yet familiar with how this code is used :smile:

src/openlifu/io/LIFUInterface.py
peterhollender: Yeah I haven't hooked the top level interface up to the lower level LIFUTxDevice. We should have a placeholder in the meantime (or I'll make the change tomorrow)

ebrahimebrahim:
@georgevigelette How are we planning to make LIFUInterface provide progress feedback? Can we have a placeholder stub thing for however that is planned?

README.rst
peterhollender: Yeah, seems reasonable to remove. I don't think I wrote that?

src/openlifu/io/LIFUUart.py
peterhollender: This may be intentional. @georgevigelette if you are trying to catch a particular exception or set of exceptions, should we catch them explicitly? This looks like on any failure (including the ValueError raised by timeout), you return an "Error" packet, rather than raising an exception.

src/openlifu/io/LIFUInterface.py
peterhollender: I believe it's a bit of both, which is not what we want. @georgevigelette and I were looking at this a little bit today. It is supposed to pretend to be connected, but the exact implementation isn't fully realized.

ebrahimebrahim: ah I see then I won't touch `test_mode` :smile: 

We could add a separate boolean argument `mock_mode` that skips the whole lower level thing, meant for use in testing the application

Or we could monkeypatch the lower level thing from the SlicerOpenLIFU end
(like after creating a `LIFUInterface` go and set its `txdevice` and so on to a custom made mock class).
Then there would be no mocking code need here in openlifu, but we would still need to have completed implementations of `set_solution`, `get_solution`, `start_sonication`, `stop_sonication`, and some progress callback/polling mechanism. Even if those implementations don't work yet due to the underlying device interfaces still being worked on, we'd still then have something to monkeypatch.

peterhollender (approved):
Looks good- we'll address any new issues in other branches

PR #215, opened 2025-02-14
Update users table to use password hashes and not passwords

### LINKED ISSUE #214: Update users table to use password hashes and not passwords
The current users table has strings and not salted bcrypt hashes of passwords, which will prevent testing login in the app, as the app will check the hashed password. Also add a `README.md` in `db_dvc/`explaining what the default credentials are, as the hashes will be unintelligible after committed to dvc.

arhowe00:
Closes #214

### DIFF
--- a/db_dvc.dvc
+++ b/db_dvc.dvc
@@ -1,6 +1,6 @@
 outs:
-- md5: bb64fad973fb4997670c236a472b96bd.dir
-  size: 803477862
-  nfiles: 82
+- md5: f9fe2e59c4e854484b4e922a719bdfa7.dir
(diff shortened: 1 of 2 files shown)


arhowe00:
Realizing that commit b97c113 might not have been necessary or accomplished anything (though it is nice to remove `Optional[str]`, I don't think this affected serialization, it is possible it did...

ebrahimebrahim (approved):
Excellent and thanks for the README!

I am curious how the `Optional[str]` gave trouble with the slicer parameter node serializer...

(either way the change is good)

PR #219, opened 2025-02-22
Fix handling of null transducer models (#218)

### LINKED ISSUE #218: `get_transducer_absolute_filepaths` cannot handle null
Since the type of both `Transducer.registration_surface_filename` and `Transducer.transducer_body_filename` are given as `Optional[str]`, these should both be allowed to be set to `None`, or equivalently `null` in the json files. However `Database.get_transducer_absolute_filepaths` will fail to handle this case because it is checking to see whether the attributes are present rather than checking to see whether the attributes are non-None:

https://github.com/OpenwaterHealth/OpenLIFU-python/blob/1dcff787b74a3b301edbab2f52f6d52835bd0198/src/openlifu/db/database.py#L709-L734

To see the bug, try `get_transducer_absolute_filepaths` on a Transducer that has either of those attributes set to `None`/`null` -- it raises an exception because it attempts to construct a `Path` using `None`.

Besides fixing this issue, it would also be nice to make `get_transducer_absolute_filepaths` just return an empty dict if there is no absolute filepath info rather than raising an exception -- it should be perfectly fine for there to be no absolute filepath info because a Transducer with `Transducer.registration_surface_filename` and `Transducer.transducer_body_filename` set to `None` is perfectly valid as things currently stand.

ebrahimebrahim:
Closes #218 

Added a test that fails before and passes after this fix

### DIFF
--- a/src/openlifu/db/database.py
+++ b/src/openlifu/db/database.py
@@ -6,7 +6,7 @@ import os
 import shutil
 from enum import Enum
 from pathlib import Path
-from typing import List, Optional
+from typing import Dict, List, Optional, Union
 
 import h5py
 
@@ -20,7 +20,7 @@ from .subject import Subject
 from .user import User
 
 OnConflictOpts = Enum('OnConflictOpts', ['ERROR', 'OVERWRITE', 'SKIP'])
-
+PathLike = Union[str, os.PathLike]
 
 class Database:
     def __init__(self, path: Optional[str] = None):
@@ -268,7 +268,13 @@ class Database:
 
         self.logger.info(f"Added subject with ID {subject_id} to the database.")
 
-    def write_transducer(self, transducer, registration_surface_model_filepath: Optional[str] = None, transducer_body_model_filepath: Optional[str] = None, on_conflict: OnConflictOpts=OnConflictOpts.ERROR) -> None:
+    def write_transducer(
+            self,
+            transducer,
+            registration_surface_model_filepath: Optional[PathLike] = None,
+            transducer_body_model_filepath: Optional[PathLike] = None,
+            on_conflict: OnConflictOpts=OnConflictOpts.ERROR,
+    ) -> None:
         """ Writes a transducer object to database and copies the affiliated transducer data files to the database if provided. When a transducer that is already present in the database is being re-written,
         the associated model data files do not need to be provided if they have previously been added to the database.
         Args:
@@ -706,7 +712,7 @@ class Database:
             return photoscan, (model_data, texture_data)
         return photoscan
 
-    def get_transducer_absolute_filepaths(self, transducer_id:str) -> dict:
+    def get_transducer_absolute_filepaths(self, transducer_id:str) -> Dict[str,Optional[str]]:
         """ Returns the absolute filepaths to the model data files i.e. transducer body and registration surface
         model files affiliated with the transducer, with ID `transducer_id`. Unlike `load_transducer`, which
         specifies the relative paths to the model datafiles along with other transducer attributes, this function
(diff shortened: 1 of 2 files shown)


ebrahimebrahim:
added @sadhana-r for visibility and but you don't have to review unless you want to
@arhowe00 please review, ty!

src/openlifu/db/database.py
```diff
@@ -707,29 +707,33 @@ def load_photoscan(self, subject_id, session_id, photoscan_id, load_data = False
         return photoscan
 
     def get_transducer_absolute_filepaths(self, transducer_id:str) -> dict:
```
arhowe00, line 709: Are you sure that this should not be returned? Something tells me we should pass along the information that nothing is associated, since this function is intended to get the absolute filepaths, you could return the key still, but with an empty string or Optional[str] with None.

tests/test_database.py
```diff
@@ -649,3 +649,38 @@ def test_write_transducer(example_database:Database, example_transducer: Transdu
...
+        registration_surface_path.touch()
+    if include_body_model:
+        transducer_body_path = Path(tmp_path/"test_db_files/example_transducer_body.obj")
+        transducer_body_path.parent.mkdir(parents=True, exist_ok=True)
+        transducer_body_path.touch()
+
+    example_database.write_transducer(
+        transducer=transducer,
+        registration_surface_model_filepath = registration_surface_path,
+        transducer_body_model_filepath = transducer_body_path,
```
arhowe00, line 672:
```diff
-        registration_surface_model_filepath = registration_surface_path,
-        transducer_body_model_filepath = transducer_body_path,
+        registration_surface_model_filepath = str(registration_surface_path),
+        transducer_body_model_filepath = str(transducer_body_path),
```

arhowe00 (changes requested):
I tested this on main and this branch with the M2 transducer, and everything seems to work. A few discussion points.

src/openlifu/db/database.py
ebrahimebrahim: empty path is probably not a good idea but I do agree with Optional[str] being a slight improvement since the type then clearly communicates to the user that they have to worry about the special case, whereas a missing key is something to worry about but the worry is not communicated in the type system

tests/test_database.py
```diff
@@ -649,3 +649,38 @@ def test_write_transducer(example_database:Database, example_transducer: Transdu
...
     with pytest.raises(ValueError, match="file associated with transducer"):
         example_database.write_transducer(example_transducer, on_conflict=OnConflictOpts.OVERWRITE)
+
+@pytest.mark.parametrize("include_registration_surface", [True, False])
+@pytest.mark.parametrize("include_body_model", [True, False])
+def test_get_transducer_absolute_filepaths(example_database:Database, tmp_path:Path, include_registration_surface:bool, include_body_model:bool):
+    transducer = Transducer(id="transducer_for_test_get_transducer_absolute_filepaths")
+
+    registration_surface_path = None
+    transducer_body_path = None
```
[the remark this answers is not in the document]
ebrahimebrahim: actually these need to be None because one of the test cases is to pass None into `write_transducer`

Is there a better way to tell the type checker what's going on? Maybe we can let it go for a unit test, but if there is a good way I am also interested

src/openlifu/db/database.py
```diff
@@ -20,7 +20,7 @@
 from .user import User
 
 OnConflictOpts = Enum('OnConflictOpts', ['ERROR', 'OVERWRITE', 'SKIP'])
-
+PathLike = Union[str, os.PathLike]
```
ebrahimebrahim, line 23: This is an attempt to make the type checker happier without fussing as much with path/str conversion when _using_ functions. LMK what you think of this. I stole the approach from MONAI: https://github.com/Project-MONAI/MONAI/blob/a09c1f08461cec3d2131fde3939ef38c3c4ad5fc/monai/config/type_definitions.py#L82

tests/test_database.py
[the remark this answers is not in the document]
arhowe00: One thing you could do is make them `Optional[str]` and derive paths from that. Gonna make that proposed change. I think it might make the intent of the tes t more clear than parametrizing with the bools

src/openlifu/db/database.py
arhowe00: Yes this is an interesting approach and it works fine!

arhowe00:
My type checker doesn't give any complaints about the one above, and we don't need to use empty path

tests/test_database.py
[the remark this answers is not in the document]
ebrahimebrahim: It looks good, thanks!

Issue #106, opened 2024-09-13
Relax linting exceptions


ebrahimebrahim:
Many linting exceptions were included when first applying the scientific-python cookie in bedc2f6e596fe262078d19a08f5bea1243e80017

Now that development on this repo is about to ramp up, it is a good time to remove some of these exceptions

ebrahimebrahim:
The linting exceptions are [here for ruff](https://github.com/OpenwaterHealth/OpenLIFU-python/blob/af3fc5741ca03d4eb87f9f10f614f5f2279efdef/pyproject.toml#L156-L190) and [here for pylint](https://github.com/OpenwaterHealth/OpenLIFU-python/blob/af3fc5741ca03d4eb87f9f10f614f5f2279efdef/pyproject.toml#L215-L247).

HussainAther:
Hey @ebrahimebrahim, I can take this on! Should we enforce a specific linting standard (e.g., PEP 8)?

ebrahimebrahim:
This issue represents a long standing effort to relax some of the ruff and pylint exceptions indicated [above](https://github.com/OpenwaterHealth/OpenLIFU-python/issues/106#issuecomment-2635547118), so I won't assign it since we have been picking away at it bit by bit. But feel free to help us pick away!

PR #224, opened 2025-02-28
STYLE: Enforce import of __future__ annotations (#106)

ebrahimebrahim:
- Remove the exception to [this linting rule](https://docs.astral.sh/ruff/rules/missing-required-import/)
- Allow ruff to make bulk changes to satisfy the new rule

Making the linter more strict on this makes it clear how to fix type annotation complaints, avoiding some confusing situations

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -5,6 +5,7 @@
 
 # -- Project information -----------------------------------------------------
 # https://www.sphinx-doc.org/en/master/usage/configuration.html#project-information
+from __future__ import annotations
 
 import importlib.metadata
 
--- a/notebooks/LIFUTestWidget.py
+++ b/notebooks/LIFUTestWidget.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import asyncio
 import logging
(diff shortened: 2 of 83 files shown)


arhowe00:
Looks good, I'd recommend modifying the commit message to also include the fact that ruff removed `Optional[•]` to use `• | None`, as I don't think it's related to `__future__` annotations, but it's fine either way and ready to rebase and merge.

arhowe00 (commented):
Looks good, I'd recommend modifying the commit message to also include the fact that ruff removed `Optional[•]` to use `• | None`, as I don't think it's related to `__future__` annotations, but it's fine either way and ready to rebase and merge.

ebrahimebrahim:
> as I don't think it's related to `__future__` annotations, but it's fine either way and ready to rebase and merge.

It is related: only once __future__ annotations become available can the `|` notation be used in python 3.9. This is why ruff then insists on doing it. standardizing will just help cause less confusion in the future

arhowe00 (approved):

PR #233, opened 2025-03-04
updated scripts for new trigger format and water tank testing

georgevigelette:
Updated test scripts for new trigger structure and added script for water tank testing.

### DIFF
--- a/notebooks/test_watertank.py
+++ b/notebooks/test_watertank.py
@@ -1,0 +1,162 @@
+from __future__ import annotations
+
+import threading
+import time
+
+import numpy as np
+
+from openlifu.bf.pulse import Pulse
+from openlifu.bf.sequence import Sequence
+from openlifu.geo import Point
+from openlifu.io.LIFUInterface import LIFUInterface
+from openlifu.plan.solution import Solution
+
+# set PYTHONPATH=%cd%\src;%PYTHONPATH%
+# python notebooks/test_watertank.py
+
+"""
+Test script to automate:
+1. Connect to the device.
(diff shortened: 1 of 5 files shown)


notebooks/test_watertank.py
```diff
@@ -0,0 +1,162 @@
...
+    sequence = sequence,
+    target=pt,
+    foci=[pt],
+    approved=True
+)
+
+sol_dict = solution.to_dict()
+profile_index = 1
+profile_increment = True
+interface.txdevice.set_solution(
```
ebrahimebrahim, line 117: could this just as well be `interface.set_solution` and just pass in the entire Solution object? or maybe that is not implemented yet

notebooks/test_watertank.py
```diff
@@ -0,0 +1,162 @@
...
+            time.sleep(log_interval)
+
+
+print("Starting LIFU Test Script...")
+interface = LIFUInterface(test_mode=False)
+tx_connected, hv_connected = interface.is_device_connected()
+if tx_connected and hv_connected:
+    print("LIFU Device Fully connected.")
+else:
+    print(f'LIFU Device NOT Fully Connected. TX: {tx_connected}, HV: {hv_connected}')
```
ebrahimebrahim, line 48: curious, would this connectivity information also eventually be contained in a call to `LIFUInterface.get_status`?

ebrahimebrahim (approved):

ebrahimebrahim:
(I will just add a reference to an issue number in the commit message before merging, so that's the only reason for my upcoming force-push)

PR #235, opened 2025-03-05
Add photocollection concept to database (#234)

### LINKED ISSUE #234: add openlifu PhotoCollection concept to database
Add the concept a collection of photos, basically just a bunch of image files of any format, living under a session.

The reason this is needed is to keep track of the the raw photo data that is collected during transducer tracking, to give it a destination after importing from device, and to link it to the Photoscan that will be generated from it.

Conceptually here is what is associated with a PhotoCollection:
- the underlying data: a bunch of files
- an associated session ID
- a unique photo collection identifier for this photo collection

So here's what'd needed:
- a `write_photocollection` function in `Database` that accepts a `List[PathLike]` pointing to some files, a subject ID, a session ID, and a photocollection ID
- a `load_photocollection_absolute_filepaths` function in `Database` that accepts subject ID, session ID, photocollection ID and returns a `List[Path]` pointing to the files.
- a `get_photocollection_ids` in `Database` that accepts subject ID, session ID and return a `List[str]` of all the photocollection IDs that are under that session.
- maybe a `write_photocollection_ids` also? but maybe not if you don't need that in a json file separately, see below.

I believe this can all be done without necessarily creating a PhotoCollection class or fiddling too much with json files...
but maybe I am being lazy and overlooking somehting

The folder structure might be something like this:
```
db
  example_subject
    example_session
      example_session.json
      photocollections
        photocollections.json  <-- just a list of IDs (but do you need this really? get_photocollection_ids could just return the folder names)
        billy_bobs_photoshoot <-- folder name is exactly the "photocollection ID" so no json file really needed. but photocollection IDs need to be sanitized so that they are appropriate to use as filenames
          img1.png
          img2.png
```

arhowe00:
Closes #234 

This is necessary for downstream work of generating a photoscan/mesh from the 70-80 photos from 3D Open Water. The 3D Open Water app **does not** generate a mesh, it just takes photos. So we store the photos for downstream analysis.

## In detail

Adds a concept of photocollections to the database. Photocollections are associated with the specific session_id of a specific subject_id, and they are identified by the reference_number. 'reference_number' was used instead of photocollection_id for consistency with 3D Open Water. 3D Open Water takes a collection of `.jpeg` photos and the goal of photocollections are to store that collection of photos into the database for downstream processing; specifically, the plan is for a photocollection of 70-80 photos of a patient's head to be used to create a mesh via an algorithm (i.e. a "photoscan" object). There is no special object associated with a photocollection -- it's just a collection of photos stored in a directory and identified by the reference_number.

## For review

This code is not unit tested but is pretty independent within the subject/session tree. Errors will likely be caught in its usage. Other than reading the code, you could pipe it through an LLM and see if you rolled the right RNG for it to discover a bug.

### DIFF
--- a/db_dvc.dvc
+++ b/db_dvc.dvc
@@ -1,6 +1,6 @@
 outs:
-- md5: cf89344a4eee1e75c5be53d91941815e.dir
-  size: 803480549
-  nfiles: 83
+- md5: 77217cb7a18172e6db4f130b4931c899.dir
+  size: 829348265
+  nfiles: 249
   hash: md5
   path: db_dvc
--- a/src/openlifu/db/database.py
+++ b/src/openlifu/db/database.py
@@ -190,6 +190,8 @@ class Database:
             self.write_solution_ids(session, [])
         if not self.get_runs_filename(subject.id, session_id).exists():
             self.write_run_ids(session.subject_id, session.id, [])
+        if not self.get_photocollections_filename(subject.id, session_id).exists():
+            self.write_reference_numbers(session.subject_id, session.id, [])
         if not self.get_photoscans_filename(subject.id, session_id).exists():
             self.write_photoscan_ids(session.subject_id, session.id, [])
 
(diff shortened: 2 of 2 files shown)


ebrahimebrahim (commented):
- Also initialize the empty folders when a new session is created, right? Is that needed?
- I think we should add the photocollection reference number as an attribute to Photoscan, so that there is a record of which photoscan came from which photocollection. But that would complicate this PR, requiring DVC data updates as well. So maybe a separate issue to be done later?

arhowe00:
> * Also initialize the empty folders when a new session is created, right? Is that needed?

Yes, that is what happens in `def write_session` when creating the empty `photocollections.json`. I might have forcepushed while you were reviewing.

> * I think we should add the photocollection reference number as an attribute to Photoscan, so that there is a record of which photoscan came from which photocollection. But that would complicate this PR, requiring DVC data updates as well. So maybe a separate issue to be done later?

I agree, I think in the end it will be helpful to also tag the photoscans, but right now we can treat them as independent and do that integration later. I'll open up an issue regarding that.

ebrahimebrahim (approved):
Looks good then, and thanks for opening that issue!

Please feel free to merge when happy with it

Issue #240, opened 2025-03-11
There is a vtk warning when displaying the skin segmentation model


sadhana-r:
After displaying the computed skin segmentation as a model node, there is a warning:

`[VTK] Bad table range: [0, -1]`

PR #243, opened 2025-03-11
Fix unnecessary scalars on surface mesh (#240)

ebrahimebrahim:
~~I think this should do it, but still need to test in slicer~~

Tested this in Slicer, it worked! No more bad table warning.


ebrahimebrahim:
@sadhana-r No need to test anything, I think this should be safe to merge

sadhana-r (approved):

Issue #242, opened 2025-03-11
SolutionAnalysis is not serialized/deserialzed to json properly


ebrahimebrahim:
After computing solution analysis, jsonification and dejsonification is not working properly, because certain attributes that are supposed to be lists of floats are coming out as numpy arrays, and the json encoder is not working with numpy arrays.

PR #244, opened 2025-03-12
Fix solution analysis result data type (#242)

ebrahimebrahim:
I believe this should fix it -- you can quickly try in SlicerOpenLIFU and see if you still get a bunch of serializer errors.

### DIFF
--- a/src/openlifu/plan/solution.py
+++ b/src/openlifu/plan/solution.py
@@ -212,7 +212,7 @@ class Solution:
             solution_analysis.p0_Pa += [np.max(p0_Pa)]
         solution_analysis.TIC = np.mean(TIC)
         solution_analysis.power_W = np.mean(power_W)
-        solution_analysis.MI = solution_analysis.mainlobe_pnp_MPa/np.sqrt(self.pulse.frequency*1e-6)
+        solution_analysis.MI = (solution_analysis.mainlobe_pnp_MPa/np.sqrt(self.pulse.frequency*1e-6)).item()
         solution_analysis.global_ispta_mWcm2 = float((ita_mWcm2*z_mask).max())
         return solution_analysis
 
--- a/tests/test_solution.py
+++ b/tests/test_solution.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from dataclasses import fields
 from datetime import datetime, timedelta
 from pathlib import Path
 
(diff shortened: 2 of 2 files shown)


arhowe00:
I've experienced some issues in SlicerOpenLIFU.  In particular, when generating the table, the traceback:

```
  File "~/SlicerOpenLIFU-Debug/lib/Slicer-5.7/qt-scripted-modules/OpenLIFUSonicationPlanner.py", line X, in populate_solution_analysis_table
    max_len = max(
ValueError: max() arg is an empty sequence
```

Here's line X:

```
    def populate_solution_analysis_table(self) -> None: # TODO: here
        ...
        # Max length of list type fields in the dataclass
        max_len = max(
            len(getattr(analysis_openlifu, f.name))
            for f in fields(analysis_openlifu)
            if get_origin(f.type) is list
        )
```



I will note that I looked at the following in the console:
```
>>> for f in fields(analysis_openlifu):
...  print(get_origin(f.type))
... 
None
None
None
None
None
None
None
None
None
None
None
None
None
None
None
None
None
None
>>> for f in fields(analysis_openlifu):
...  print(f.type)
... 
list[float]
list[float]
list[float]
list[float]
list[float]
list[float]
list[float]
list[float]
list[float]
list[float]
list[float]
list[float]
list[float]
list[float]
float | None
float | None
float | None
float | None
```

arhowe00 (changes requested):
See above about whether I should change something in SlicerOpenLIFU or this should be addressed here. `get_origin(f.type)` could be returning `None` because `f.type` is not a generic alias but instead a raw type annotation from `dataclasses.fields()`.  In the docs for `get_origin`:

```
(function) def get_origin(tp: Any) -> (Any | None)
──────────────────────────────────────────────────────────────────────────────
Get the unsubscripted version of a type.                                      

This supports generic types, Callable, Tuple, Union, Literal, Final, ClassVar,
Annotated, and others. Return None for unsupported types.

Examples:
    >>> P = ParamSpec('P')
    >>> assert get_origin(Literal[42]) is Literal
    >>> assert get_origin(int) is None
    >>> assert get_origin(ClassVar[int]) is ClassVar
    >>> assert get_origin(Generic) is Generic
    >>> assert get_origin(Generic[T]) is Generic
    >>> assert get_origin(Union[T, int]) is Union                             
    >>> assert get_origin(List[Tuple[T, T]][int]) is list
    >>> assert get_origin(P.args) is P
```
i.e. for this definition, the type annotation is returning none because it's not actually using Typing "`List`"

```
..
    mainlobe_pnp_MPa: list[float] = field(default_factory=list)
    mainlobe_isppa_Wcm2: list[float] = field(default_factory=list)
    global_isppa_Wcm2: list[float] = field(default_factory=list)
...
```

ebrahimebrahim:
Hmm but in a python console (including in slicer's python version), this works:

```
from typing import get_origin
get_origin(list[float])
```

it returns float

arhowe00:
> > the type annotation is returning none because it's not actually using Typing "`List`"
> 
> Hmm but in a python console (including in slicer's python version), this works:
> 
> ```
> from typing import get_origin
> get_origin(list[float])
> ```
> 
> it returns ~float~ `list`

Right, but this isn't exactly the same as doing `get_origin(f.type)`, as above. The following _does_ work:

```
>>> for f in fields(analysis_openlifu):
...     if str(f.type) == "list[float]":
...         print(f.type)
list[float]
...
```
So I'll do that in addressing [S#206](https://github.com/OpenwaterHealth/SlicerOpenLIFU/issues/206)

arhowe00:
Okay, everything works. Please rebase and merge

PR #247, opened 2025-03-14
Lower level hardware interface updates 

georgevigelette:
Updates to facilitate testing and enabling multiple transmitter modules


tests/test_sonication_control_mock.py
```diff
@@ -33,7 +33,7 @@ def example_solution() -> Solution:
 
 def test_lifuinterface_mock(example_solution:Solution):
     """Test that LIFUInterface can be used in mock mode (i.e. test_mode=True)"""
-    lifu_interface = LIFUInterface(test_mode=True)
+    lifu_interface = LIFUInterface(TX_test_mode=True, HV_test_mode=True)
```
ebrahimebrahim, line 36: @arhowe00 Note the change here to the test mode arguments in `LIFUInterface`

The update will need to be propagated to SlicerOpenLIFU

M4_flex.json
ebrahimebrahim, line 1: Should remove this empty file here, probably got committed by accident

ebrahimebrahim (changes requested):

M4_flex.json
ebrahimebrahim: I will do it now

tests/test_sonication_control_mock.py
georgevigelette: yes this was done so that you can put either device in test mode, more important for water tank testing where a lab supply may be in use and want to put the High Voltage console in test mode.

M4_flex.json
georgevigelette: Yes it was suppose to be for 2 module pin mapping never put it in there. it can be removed.

Issue #248, opened 2025-03-17
Add method of switching LIFUInterface in/out of test mode


arhowe00:
Add a method to switch LIFUInterface between test and live mode at runtime since only one instance should be instantiated with the SlicerOpenLIFU application. This ensures the interface can start in live mode for monitoring USB connections but later transition to test mode if a developer user decides to mock the hardware interface, without creating a new instance.

ebrahimebrahim:
Closing as not planned, because in a discussion with George a while back the conclusion was that it's too hard to make test mode be a dynamically changeable thing, and George recommended that we reconstruct LIFUInterface whenever we want to change the test mode.  This is exactly what you ended up doing in SlicerOpenLIFU.

PR #140, opened 2024-10-23
Port over minimal solution analysis code from matlab

ltetrel:
This PR re-opens the folowing that was accidentally merged: https://github.com/OpenwaterHealth/OpenLIFU-python/pull/135

PR #249, opened 2025-03-17
Add `toggle_test_mode` to LIFUInterface (#248)

arhowe00:
## For review

Please check that this function handles everything necessary to set the LIFUInterface in test_mode (I just reinitialize some dependent objects and reconnect signals. Note that the old device objects are orphaned).

@georgevigelette

### DIFF
--- a/src/openlifu/io/LIFUInterface.py
+++ b/src/openlifu/io/LIFUInterface.py
@@ -213,6 +213,34 @@ class LIFUInterface:
             logger.error("Error Stopping sonication: %s", e)
             raise e
 
+    def toggle_test_mode(self, enabled : bool) -> None:
+        """
+        Toggle test mode and reinitialize interface components.
+
+        Args:
+            enabled (bool): If True, enables test mode; otherwise, disables it.
+        """
+        self._test_mode = enabled
+
+        # Create a TXDevice instance as part of the interface
+        logger.debug("Re-initializing TX Module of LIFUInterface with VID: %s, PID: %s, baudrate: %s, timeout: %s", self.vid, self.tx_pid, self.baudrate, self.timeout)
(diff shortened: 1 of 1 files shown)


ebrahimebrahim (commented):
It looks good to me; I think @georgevigelette should check that this covers everything needed to toggle testing mode for both Tx and HV. Like the switching of `demo_mode` on `LIFUUart` @georgevigelette should confirm this way of doing it works. Thanks!

arhowe00:
Reminder for myself to edit ` python-requirements.txt` in `Change LIFUInterface to use test mode toggle (#140)` if the hash changes upon merge.

georgevigelette:
@arhowe00  I would agree, probably shouldn't need to recreate the instance.  I was out yesterday so I will look at this now and get it approved.

georgevigelette:
I think this may be enough and just need to update the UART to handle if there is physical uart connection and you toggle demo mode and then back to live handle the closing and opening of the COM port.
In LIFUInterface.py
    def toggle_test_mode(self, enabled : bool) -> None:
        """
        Toggle test mode and reinitialize interface components.

        Args:
            enabled (bool): If True, enables test mode; otherwise, disables it.
        """
        self._test_mode = enabled

        # Switch TXDevice into demo mode
        if self._tx_uart:
            self._tx_uart.set_demo_mode(enabled)

        # Switch ControllerDevice into demo mode       
        if self._hv_uart:
            self._hv_uart.set_demo_mode(enabled)

In the LIFUUart.py (need to test to see what needs to happen in switching back and forth from live/demo mode)
    def set_demo_mode(self, demo_mode: bool):
        """Set the demo mode flag."""
        self.demo_mode = demo_mode
        if self.demo_mode:
            log.info("Demo mode enabled.")
        else:
            log.info("Demo mode disabled.")

ebrahimebrahim:
@arhowe00 Once you confirm that destroying and recreating LIFUInterface can work with the Slicer event loop, you can close this PR

Issue #251, opened 2025-03-19
Integrate meshroom workflow


ebrahimebrahim:
Integrate a meshroom workflow that, assuming meshroom is installed, calls a subprocess to turn a `List[Path]` photocollection into  a `Photoscan`.

For now it's okay to just take one of the several generated texture files. We will address the issue of multiple texture files in follow-up work

PR #252, opened 2025-03-19
Basic Meshroom Integration (Related to #251)

dlichy:
Added a function to run Meshroom with default pipeline and return a Photoscan object. Also added a meshroom_pipelines folder in src directory to hold Meshroom configurations.

### DIFF
--- a/src/openlifu/meshroom_pipelines/default_pipeline.mg
+++ b/src/openlifu/meshroom_pipelines/default_pipeline.mg
@@ -1,0 +1,180 @@
+{
+    "header": {
+        "pipelineVersion": "2.2",
+        "releaseVersion": "2023.3.0",
+        "fileVersion": "1.1",
+        "template": true,
+        "nodesVersions": {
+            "MeshFiltering": "3.0",
+            "DepthMapFilter": "4.0",
+            "MeshDecimate": "1.0",
+            "Meshing": "7.0",
+            "FeatureExtraction": "1.3",
+            "CameraInit": "9.0",
+            "Texturing": "6.0",
+            "ImageMatching": "2.0",
+            "Publish": "1.3",
+            "DepthMap": "5.0",
+            "FeatureMatching": "2.0",
+            "PrepareDenseScene": "3.1",
+            "StructureFromMotion": "3.3"
+        }
+    },
+    "graph": {
+        "Meshing_1": {
+            "nodeType": "Meshing",
+            "position": [
+                1600,
+                0
+            ],
+            "inputs": {
+                "input": "{DepthMapFilter_1.input}",
+                "depthMapsFolder": "{DepthMapFilter_1.output}"
+            }
+        },
+        "DepthMapFilter_1": {
+            "nodeType": "DepthMapFilter",
+            "position": [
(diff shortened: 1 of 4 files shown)


src/openlifu/meshroom_pipelines/default_pipeline.mg
ebrahimebrahim, line 1: When you pip install openlifu the wheel would probably not include the `default_pipeline.mg` because it doesn't look like a python source file. I think it needs to be added to the manifest by adding something like [this](https://setuptools.pypa.io/en/latest/userguide/datafiles.html#package-data):

```
[tool.setuptools.package-data]
openlifu = ["meshroom_pipelines/*.mg"]
```

I did not test any of that but please give it a try: try a fresh clean (non-editable) pip install from this branch and make sure `default_pipeline.mg` makes it into the installed `site_packages/blah/blah/...`

src/openlifu/photoscan.py
```diff
@@ -207,3 +210,48 @@ def convert_between_ras_and_lps(mesh : vtk.vtkPointSet) -> vtk.vtkPointSet:
     transformFilter.Update()
 
     return transformFilter.GetOutput()
+
+def run_reconstruction(images: list[Path],
+                       pipeline: Path = Path(__file__).resolve().parent / "meshroom_pipelines" / "default_pipeline.mg") -> Photoscan:
```
ebrahimebrahim, line 215: Can we
- make `pipeline` default to `None` and have type `Optional[Path]`. Then inside the function you see if it's `None` and set it to the right thing if so.
- instead of messing with `__file__` here, use [importlib](https://docs.python.org/3/library/importlib.resources.html#importlib.resources.path): `importlib.resources.path("openlifu.meshroom_pipelines", "default_pipeline.mg")` to get the path

ebrahimebrahim (changes requested):

ebrahimebrahim:
@arhowe00 would just like your check of the overall design here to make sure it makes sense. i.e. it integrates appropriately with your photocollection thing, and it provides what's needed for you to call from SlicerOpenLIFU and to write the generated `Photoscan` to `Database` in the case of a workflow that uses the database

ebrahimebrahim (commented):
Ah also we need an update to the README that adds the needed setup steps for windows at least

ebrahimebrahim:
@dlichy what OS did this work on for you? ubuntu? do you have access to a windows machine for testing?

arhowe00:
Working on this at the moment

dlichy:
> @dlichy what OS did this work on for you? ubuntu? do you have access to a windows machine for testing?

Yes Ubuntu. I do not have access to a windows machine.

ebrahimebrahim:
@dlichy Can you add those setup steps to the README?

Then @arhowe00 can do a final basic code review

ebrahimebrahim:
Thanks the README looks good, we can add windows instructions once we figure out what they should actually be!

@arhowe00 this is ready for a code-only review from you

(in parallel you can also test this out on windows when you get a chance and see how that goes!)

ebrahimebrahim (approved):

PR #254, opened 2025-03-20
Add virtual fit options to protocol and transducer

### LINKED ISSUE #165: Add virtual fit options to Protocol
> pitch and yaw ranges should come from the Protocol, as they are determined by the target's general location. I think it's ok for virtual fitting to take them as inputs (as opposed to taking the whole Protocol), but their values in operation should come from the Protocol.

_Originally posted by @peterhollender in https://github.com/OpenwaterHealth/OpenLIFU-python/pull/162#discussion_r1848790409_

Add virtual fit options (perhaps VirtualFitOptions similar to SolutionAnalysisOptions) to Protocols

They would cover:

- transducer_steering_center_distance: Distance from the transducer origin axially to the center of the steering zone, in the units of `volume_affine_RAS`.
- steering_limits: Left and right extents of the steering zone in the lateral, elevational, and then axial directions, measured from the transducer steering center.
- pitch_range: Range of pitches to include in the transducer fitting search grid, in degrees
- pitch_step: Pitch step size when forming the transducer fitting search grid, in degrees
- yaw_range: Range of yaws to include in the transducer fitting search grid, in degrees
- yaw_step: Yaw step size when forming the transducer fitting search grid, in degrees
- planefit_dyaw_extent: Left and right extents of the point grid to be used for plane fitting along the local yaw axes, in spatial units of `volume_affine_RAS`. The plane fitting point grid will be twice this size, since this is left and right extents. (Note that this has units of length, not angle!)
- planefit_dyaw_step: Local yaw axis step size to use when constructing plane fitting grids. In spatial units of `volume_affine_RAS`.
- planefit_dpitch_extent: Left and right extents of the point grid to be used for plane fitting along the local pitch axes, in spatial units of `volume_affine_RAS`. The plane fitting point grid will be twice this size, since this is left and right extents.
- planefit_dpitch_step: Local pitch axis step size to use when constructing plane fitting grids. In spatial units of `volume_affine_RAS`.

Depends on #162

---

Now that virtual fit has been refactored, here are the things that ought to be part of protocol:
- the steering ranges (they are transducer-specific but still could vary by protocol)
- the virtual fitting search grid
- maybe the planefit parameters

### LINKED ISSUE #166: Add standoff transform to Transducer definition
Add `standoff_transform`, a concept that is to be added when the #147 is done, to Transducer definition.

ebrahimebrahim:
Close #165 
Close #166 

- Package virtual fit options into a VirtualFitOptions dataclass
- Add that VirtualFitOptions to Protocol
- Update test database accordingly
- Move ArrayTransform into geo. It's about time we make that available outside of database related stuff. Initially did this to use ArrayTransform
- Build array conversion into DictMixin (saves some effort whenever we create a new array type field in a dataclass, so we don't have to manually fiddle with the to_dict and from_dict to convert things to numpy arrays)
- Add standoff transform to Transducer


ebrahimebrahim:
@sadhana-r No need to check anything here, you can test it out in https://github.com/OpenwaterHealth/SlicerOpenLIFU/pull/212 instead and approve this if you approve of that

sadhana-r (approved):

PR #259, opened 2025-03-26
Add heterogenous param extraction

### LINKED ISSUE #258: k-Wave simulation ignore heterogenous params
When `get_medium` was first implemented, it apparently ignored the actual values of the params Dataset, and only used the `ref_value` field of the `params.attrs`. Extraction of the actual parameter arrays for creating the medium needs to be implemented

peterhollender:
Closes #258

### DIFF
--- a/src/openlifu/sim/kwave_if.py
+++ b/src/openlifu/sim/kwave_if.py
@@ -49,9 +49,19 @@ def get_karray(arr: xdc.Transducer,
     karray.set_array_position(translation, rotation)
     return karray
 
-def get_medium(params: xa.Dataset):
-    medium= kWaveMedium(sound_speed=params['sound_speed'].attrs['ref_value'],
-                       density=params['density'].attrs['ref_value'])
+def get_medium(params: xa.Dataset, ref_values_only: bool = False):
+    if ref_values_only:
+        medium = kWaveMedium(sound_speed=params['sound_speed'].attrs['ref_value'],
+                             density=params['density'].attrs['ref_value'],
+                             alpha_coeff=params['attenuation'].attrs['ref_value'],
+                             alpha_power=0.9,
+                             alpha_mode='no_dispersion')
+    else:
+        medium= kWaveMedium(sound_speed=params['sound_speed'].data,
+                        density=params['density'].data,
+                        alpha_coeff=params['attenuation'].data,
+                        alpha_power=0.9,
+                        alpha_mode='no_dispersion')
     return medium
 
 def get_sensor(kgrid, record=['p_max','p_min']):
@@ -79,7 +89,8 @@ def run_simulation(arr: xdc.Transducer,
                    cfl: float = 0.5,
                    bli_tolerance: float = 0.05,
                    upsampling_rate: int = 5,
-                   gpu: bool = True
+                   gpu: bool = True,
+                   ref_values_only: bool = False
 ):
     delays = delays if delays is not None else np.zeros(arr.numelements())
     apod = apod if apod is not None else np.ones(arr.numelements())
@@ -94,7 +105,7 @@ def run_simulation(arr: xdc.Transducer,
                         translation=array_offset,
                         bli_tolerance=bli_tolerance,
                         upsampling_rate=upsampling_rate)
-    medium = get_medium(params)
+    medium = get_medium(params, ref_values_only=ref_values_only)
     sensor = get_sensor(kgrid, record=['p_max', 'p_min'])
     source = get_source(kgrid, karray, source_mat)
     logging.info("Running simulation")
--- a/tests/test_sim.py
+++ b/tests/test_sim.py
@@ -1,10 +1,20 @@
 from __future__ import annotations
 
+import sys
+
+import pytest
 import xarray
 
 import openlifu
 
 
+@pytest.mark.skipif(
+    sys.platform == 'darwin',
+    reason=(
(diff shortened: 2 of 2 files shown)


peterhollender:
@ebrahimebrahim I'm not sure why CI is failing on macOS with an overflow error - does that CI result mean something to you?

arhowe00:
I also got a failure here. However, I'm using Python 3.12.  It threw within the logger, I can investigate and debug tomorrow:

```pytest
self = <LogRecord: root, 20, /Users/andrew.howe/projects/ow/OpenLIFU-python/env-3.12.2/lib/python3.12/site-packages/kwave/kWaveSimulation_helper/set_sound_speed_ref.py, 89, "  reference sound speed: ">

    def getMessage(self):
        """
        Return the message for this LogRecord.

        Return the message for this LogRecord after merging any user-supplied
        arguments with the message.
        """
        msg = str(self.msg)
        if self.args:
>           msg = msg % self.args
E           TypeError: not all arguments converted during string formatting

msg        = '  reference sound speed: '
self       = <LogRecord: root, 20, /Users/andrew.howe/projects/ow/OpenLIFU-python/env-3.12.2/lib/python3.12/site-packages/kwave/kWaveSimulation_helper/set_sound_speed_ref.py, 89, "  reference sound speed: ">

../../../.pyenv/versions/3.12.2/lib/python3.12/logging/__init__.py:392: TypeError
...
FAILED tests/test_sim.py::test_run_simulation_runs - TypeError: not all arguments converted during string formatting
```

peterhollender:
> It's probably best first to debug the error in the CI, as my installation has a few workarounds.
> 
> So apparently squaring `output['p_min'].reshape(sz, order='F')` is causing overflow. float64 overflows at 1e308. For reference, this happens here:
> 
> ```python
> >       intensity = xa.DataArray(1e-4*output['p_min'].reshape(sz, order='F')**2/(2*Z),
>                              coords=params.coords,
>                              name='I',
>                              attrs={'units':'W/cm^2', 'long_name':'Intensity'})
> ```
> 
> and some of these values for p_min look very large, like -3.4028235e+38 will square to 1e77, I assume there may be very large values in an array.
> 
> ```python
> output     = {'Nt': array(3, dtype=uint64), 'Nx': array(81, dtype=uint64), 'Ny': array(81, dtype=uint64), 'Nz': array(81, dtype=uint64), ...}
> p_max      = <xarray.DataArray 'p_max' (lat: 21, ele: 21, ax: 13)> Size: 23kB
> array([[[-3.4028235e+38, -3.4028235e+38, -3.4028235e+...    (ax) float64 104B -2.0 -1.0 0.0 1.0 2.0 ... 6.0 7.0 8.0 9.0 10.0
> Attributes:
>     units:      Pa
>     long_name:  PPP
> p_min      = <xarray.DataArray 'p_min' (lat: 21, ele: 21, ax: 13)> Size: 23kB
> array([[[-3.4028235e+38, -3.4028235e+38, -3.4028235e+...    (ax) float64 104B -2.0 -1.0 0.0 1.0 2.0 ... 6.0 7.0 8.0 9.0 10.0
> Attributes:
>     units:      Pa
>     long_name:  PNP
> ```

Wow, this is not at all what I get on my machine...
p_min.min() is -3194.65 and p_min.max() is 0.0, so intensity.max() is just .00034019. 

I got the values by debugging `test_sim.test_run_simulation_runs()` and putting a breakpoint in `kwave_if.py` after the `intensity` calculation.

peterhollender:
> @peterhollender We have a minimal test here that just checks that the simulation runs at all, to help catch issues we've had in the past with a dependency update breaking kwave on one platform or python version:
> 
> https://github.com/OpenwaterHealth/OpenLIFU-python/blob/main/tests/test_sim.py
> 
> When you look at that test do you see anything stand out as being unreasonable parameters that could cause overflow?

Values look fine on Windows. It's certainly a reductive simulation which could be like... falling apart (if you undersample sims, sometimes it goes haywire at boundaries), but I don't know why we'd see that only on mac, while windows remains well-behaved

ebrahimebrahim:
I figured out the logger bug that we were getting when running `pytest -k test_sim`: https://github.com/OpenwaterHealth/OpenLIFU-python/issues/328

peterhollender:
> Peter we've looked at the macos overflow issue and narrowed it down to the kwave binary doing different things on different systems.
> 
> In all cases, the input file that gets delivered into the kwave binary looks like this (this is a `print` so arrays can get elided, but hopefully enough of the idea is there to understand):
> 
> input to kwave
> However the outputs are different on my linux system versus Andrew's mac system. Linux:
> 
> output from kwave (linux)
> Mac:
> 
> output from kwave (mac)
> So exactly the same input to the binary, but two different outputs. You can see the unreasonable values for p_max and p_min in the mac output.
> 
> We tried a few things like adding more time steps and pushing the target out further, but we still get the overflow on mac.
> 
> @peterhollender When you look at the raw input to kwave above, does anything stand out to you as causing an unreasonable simulation?
Not really, but it's a very reductive simulation, so maybe something is coming up as zeros. Tried updating some of the sim parameter to explicitly use zero-delays.

ebrahimebrahim:
> Have we seen normal output from them on mac with full-sized sims?

Yes, we have! The normal simulations work just fine.

I'll get this merged soon, skipping `test_sim` if I have to

ebrahimebrahim:
Heads up that I am going to rebase this

Issue #269, opened 2025-04-02
meshroom pipelines resource not importable


ebrahimebrahim:
Doing something like

```py
import importlib.resources
importlib.resources.path("openlifu.meshroom_pipelines", "default_pipeline.mg")
```

fails, ~~but I think it only fails when you have a non-editable install so this is tricky to catch.~~

This failure causes the mesh reconstruction to fail with a non-editable openlifu-python install.

ebrahimebrahim:
Further subtlety to notice this bug: it only happens in python 3.9, not in 3.10 onward

PR #270, opened 2025-04-02
Fix meshroom pipeline resource import in python 3.9

ebrahimebrahim:
Fixes the bug #269

The fix is simply to add an `__init__` file


ebrahimebrahim:
(@arhowe00  added a unit test that I confirmed fails in a python3.9 environment before this fix, so you can just approve if you trust the unit test)

ebrahimebrahim:
Uhh I wasn't expecting that test to to fail...
I will re-request review after I deal with this...

arhowe00 (approved):
Looks good to merge.

PR #276, opened 2025-04-10
Enable High Voltage 

georgevigelette:
High Voltage enable and disable added to sonication start stop to facilitate sonication.


environment_pyfus.yml
ebrahimebrahim, line 1: May I remove this top level `environment_pyfus.yml` file before squashing-and-merging to main?

Instead of freezing an environment like this you can pip install the package. Version requirements and everything are specified in the `pyproject.toml` file, and if there is something that needs to be a particular version for everything to work then we should specify that there. There is an advantage to you doing your testing in the same environment that a user would be in when they go and install the software: it can help catch dependency version issues!

When developing I recommend to `pip install` with the `-e` option so you can edit the codebase freely and have that immediately update the installed package

georgevigelette: Yes you can remove it, I don't have anything that relies on specific versions of libraries.

ebrahimebrahim (approved):

PR #280, opened 2025-04-16
Add user roles to individual protocols

### LINKED ISSUE #239: Add user roles to individual protocols
For some protocols we may we want to limit the usage based on user roles. We can do this by allowing a Protocol to specify as one of its attributes a list of user roles.

Once this is done, we can open an issue on SlicerOpenLIFU to work on enforcing the user role when in "user account mode". Enforcing it could mean anything and is up to the application, but one idea is, whenever showing protocols to choose from, to simply filter out and not show any protocols that have this user role attribute and for which the currently logged in user does not match. Something would also have to be done for loading protocols from file.

arhowe00:
Closes #239 

Adds a field to the protocol class that holds a list of roles, so that individual protocols can be designated for use by specific groups.

- [x] TODO: Once this is done, we can open an issue on SlicerOpenLIFU to work on enforcing the user role when in "user account mode". Enforcing it could mean anything and is up to the application, but one idea is, whenever showing protocols to choose from, to simply filter out and not show any protocols that have this user role attribute and for which the currently logged in user does not match. Something would also have to be done for loading protocols from file.

### DIFF
--- a/src/openlifu/plan/protocol.py
+++ b/src/openlifu/plan/protocol.py
@@ -41,6 +41,9 @@ class Protocol:
     description: Annotated[str, OpenLIFUFieldData("Protocol description", "A more detailed description of the protocol")] = ""
     """A more detailed description of the protocol"""
 
+    allowed_roles: Annotated[List[str], OpenLIFUFieldData("Allowed roles", "A list of user roles allowed to interact with this protocol")] = field(default_factory=list)
+    """A list of user roles allowed to interact with this protocol"""
+
     pulse: Annotated[bf.Pulse, OpenLIFUFieldData("Pulse definition", "The pulse definition used in the protocol")] = field(default_factory=bf.Pulse)
     """The pulse definition used in the protocol"""
 


ebrahimebrahim:
Make sure to double check with Peter that this is more or less the model of user-protocol permission relations that he had in mind

sadhana-r:
@arhowe00 What did you change in db_dvc? I don't see an `allowed_roles` field in any of the protocols.

arhowe00:
> @arhowe00 What did you change in db_dvc? I don't see an `allowed_roles` field in any of the protocols.

Good point. I did run the script and it did not add an empty list to protocols. However, I was able to load the protocol from Database. My guess is that by default, empty lists are not serialized, and a lack of a field for `allowed_roles` populates the Python object with a field with the list as empty.

> And should all protocols default to include atleast Administrator (or whatever the highest user role is)? Or will an empty list mean that everyone has access?

This is to be decided, but I think it's better to specify less unless required.  So an empty list means that, if user roles are enforced, nobody has access except for admins. Administrators have access to all data all the time. So if you aren't an admin (e.g. you have the role `custom_operator_role`), you cannot access the protocol unless your role is added to the protocol.

sadhana-r (approved):

arhowe00:
@ebrahimebrahim I checked off the TODO because you opened this issue in SlicerOpenLIFU (i.e. I did not do anything that I marked TODO). https://github.com/OpenwaterHealth/SlicerOpenLIFU/issues/262

Issue #279, opened 2025-04-15
Redirect meshroom logs when calling `run_reconstruction`


ebrahimebrahim:
Currently the meshroom process writes its progress to stdout.

(1) When using subprocess we should take that stdout and direct it to a [logger](https://docs.python.org/3/library/logging.html#logger-objects) named "meshroom" or something like that. Applications can then decide what to do with those logs.

stdout can be logged at the info level and stderr can be logged at the error level

(2) There is also log output that is saved by meshroom -- if that's any different or more detailed then it may be handy to somehow return that as an object from `run_reconstruction`.

It could be optional, like only if `return_logs=True` is set in `run_reconstruction` would it return the logs in some form (a giant string? or a dict of strings if there are many log files? or file objects that point to open log files in the tmp workspace so the user can read them if they want to?)

ebrahimebrahim:
Closed via #281

PR #281, opened 2025-04-21
Add Meshroom logging to run_reconstruction (#279)

dlichy:
Add functionality to redirect Meshroom logs to the python logger.


ebrahimebrahim (changes requested):
Nice, this will be super useful in SlicerOpenLIFU!

The `CalledProcessError` is also great because we can catch that in SlicerOpenLIFU and handle meshroom failures more gracefully

I like that the MeshRecon logger is separate from the Meshroom logger.

Small change requested

ebrahimebrahim (approved):
Looks perfect, thank you!

PR #289, opened 2025-04-24
P/segmethod cleanup

peterhollender:
This PR refactors the way SegmentationMethods work

## Before merging

- [x] Do a merge commit of the `db_dvc` from main and this branch.


src/openlifu/seg/seg_methods/uniform.py
```diff
@@ -0,0 +1,35 @@
...
+        return self._ref_segment(vol.coords)
+
+class UniformTissue(UniformSegmentation):
+    """ Assigns the tissue material to all voxels in the volume. """
+    def __init__(self, materials=None):
+        super().__init__(materials=materials, ref_material="tissue")
+
+    def to_dict(self):
+        d = super().to_dict()
+        d.pop("ref_material")
```
arhowe00, line 19: Why are you popping `ref_material` here? Isn't it important?

src/openlifu/seg/seg_methods/uniform.py
```diff
@@ -0,0 +1,35 @@
...
+        d["class"] = "UniformTissue"
```
arhowe00, line 20: I think this doesn't need to be set because the parent class handles this.

src/openlifu/seg/seg_methods/uniform.py
```diff
@@ -0,0 +1,35 @@
...
+
+class UniformWater(UniformSegmentation):
+    """ Assigns the water material to all voxels in the volume. """
+    def __init__(self, materials=None):
+        super().__init__(materials=materials, ref_material="water")
+
+    def to_dict(self):
+        d = super().to_dict()
+        d.pop("ref_material")
+        d["class"] = "UniformWater"
```
arhowe00, line 31: Same points as above

arhowe00 (changes requested):
This is looking great! I made some changes to simplify the base class and added type annotations, which have been incredibly helpful. I also added tests for `Material` and additional tests for `SegmentationMethod`.

I didn't make changes on more substantive questions I had, so I left those as comments.

When you look at these, I'll see if there are any more tests I can add so this integrates with the app

src/openlifu/seg/seg_methods/uniform.py
```diff
@@ -0,0 +1,32 @@
+from __future__ import annotations
+
+import xarray as xa
+
+from openlifu.seg.seg_method import SegmentationMethod
+
+
+class UniformSegmentation(SegmentationMethod):
```
arhowe00, line 8: About the inheritance hierarchy:

$$SegmentationMethod > UniformSegmentation > UniformTissue == UniformWater$$

Since `UniformSegmentation` only defines `_segment` which the others use, is it okay to just repeat the `_segment` method for both of them?  What is the motivation behind the 3 layers of inheritance?

arhowe00 (commented):
Additional question I noticed when testing things out.

src/openlifu/seg/seg_methods/uniform.py
arhowe00: It also might make configuration of SegmentationMethod for protocols a bit confusing -- but absolutely doable. 

For instance, if wanting to select a segmentation method for protocol, a researcher would first have to decide on the **middle** class in the hierarchy (if you add equals to `UniformSegmentation`), and then they would have to choose the **third** class in the hierarchy (in this case it would just be `UniformTissue` or `UniformWater`).  

Right now, the configuration should be pretty straightforward, but if you do intend there to be alternatives to `UniformSegmentation` and then alternatives in the 3rd layer within that selection, I'll add the logic for that in the configuration panel.

src/openlifu/seg/seg_methods/uniform.py
peterhollender: `ref_material` is not a valid input for `UniformTissue()` as it is fixed to be "tissue" on instanstantiation.

src/openlifu/seg/seg_methods/uniform.py
peterhollender: Good call. Removing

src/openlifu/seg/seg_methods/uniform.py
peterhollender: same

src/openlifu/seg/seg_methods/uniform.py
peterhollender: Options are `UniformSegmentation` (with ability to configure materials and pick ref_material), `UniformWater` (with ability to configure materials) and `UniformTissue` (with ability to configure materials). The fact that `UniformWater` and `UniformTissue` inherit from `UniformSegmentation` is not relevant to the user.

src/openlifu/seg/seg_methods/uniform.py
arhowe00: We can assume it will always be "tissue" for `UniformTissue`, but we should still keep it as an attribute of the class, as it technically is (and is mutable). If we want the reference materials to be embedded in the class names, that notion should be final and `ref_material` shouldn't be an attribute of `SegmentationMethod`. I think adding it back makes no difference

arhowe00 (approved):
Looks good. I did test this branch in the extension, and the previous issues with using segmentation methods went away.

PR #295, opened 2025-04-30
Serialize allowed-roles in `Protocol.to_dict` (#294)

### LINKED ISSUE #294: Protocols don't serialize allowed_roles
@sadhana-r noticed this would potentially be a bug in #239 , and [my initial guess](https://github.com/OpenwaterHealth/OpenLIFU-python/pull/280#issuecomment-2819031771) about the empty list not being serialized was wrong. It turns out, the following change would be needed to make sure the Protocol can be serialized

```diff
diff --git a/src/openlifu/plan/protocol.py b/src/openlifu/plan/protocol.py
index 40bab26..8470909 100644
--- a/src/openlifu/plan/protocol.py
+++ b/src/openlifu/plan/protocol.py
@@ -108,6 +108,7 @@ class Protocol:
             "id": self.id,
             "name": self.name,
             "description": self.description,
+            "allowed_roles": self.allowed_roles,
             "pulse": self.pulse.to_dict(),
             "sequence": self.sequence.to_dict(),
             "focal_pattern": self.focal_pattern.to_dict(),
```

### LINKED ISSUE #296: `param_constraints` in `SolutionAnalysisOptions` Not Reconstructed on JSON Deserialization
## Summary

An issue with SolutionAnalysisOptions serialization leads to replacing ParameterConstraint objects with dicts instead of `ParameterConstraint` objects.  This bug was discovered in [SlicerOpenLIFU#284](https://github.com/OpenwaterHealth/SlicerOpenLIFU/pull/284#issuecomment-2838936531)

In integrating OpenLIFU with SlicerOpenLIFU, we’ve identified a bug where `param_constraints` from a `SolutionAnalysis` instance become plain dictionaries instead of `ParameterConstraint` instances. This happened through a chain of serialization/deserialization steps—likely when the `SolutionAnalysisOptions` is serialized to JSON and then read back. The symptom is that accessing methods like `is_warning()` on `param_constraints` fails with an `AttributeError`, confirming the object is no longer a true `ParameterConstraint`.

## Repro Notes:
 - Happens when using compute_solution_openlifu(...) from the Slicer extension.
 - Reproducible with strict constraint on example_protocol.param_constraints, which fails due to a dict being returned in place of a ParameterConstraint.
 - Confirmed that protocol JSON loading correctly reconstructs ParameterConstraint objects; failure seems isolated to SolutionAnalysisOptions.

### LINKED ISSUE #298: Parameter constraints not initializing with proper values
When initializing parameter constraints that involve double-sided bounds, despite having valid constraints (warning bounds are tighter than error bounds), an exception is thrown. For instance:

```python
import openlifu
constraint = openlifu.plan.ParameterConstraint('within', (1.0, 2.0), (0.0, 4.0))
```

yields the following traceback:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "<string>", line 6, in __init__
  File "src/openlifu/plan/param_constraint.py", line 33, in __post_init__
    raise ValueError("Error value must be a sorted tuple of two numbers")
ValueError: Error value must be a sorted tuple of two numbers
```

arhowe00:
Closes #294 
Closes #296 
Closes #298
Contributes to #297 

- Fixes a small issue where allowed_roles could not be serialized.
- Also update db_dvc after running the script in `scripts/standardize_database.py`, which now shows the serialized `Protocol.allowed_roles`
- Fixes an issue where SolutionAnalysis and SolutionAnalysisOptions were not serialized correctly. It does this by introducing failing tests and fixing them.
- Fixes an issue where `ParameterConstraint` had an incorrect `__post_init__` (and adds tests)

## For review

Check the code is good enough. I checked out this branch in the extension and was able to fully interact with Protocols without issues (saving and loading).  Furthermore, after using this branch, the error in [SlicerOpenLIFU#284](https://github.com/OpenwaterHealth/SlicerOpenLIFU/pull/284) regarding the ParameterConstraint objects being swapped for dictionaries went away.

### DIFF
--- a/tests/test_protocol.py
+++ b/tests/test_protocol.py
@@ -31,6 +31,11 @@ def example_session() -> Session:
 def example_wheel_pattern() -> Wheel:
     return Wheel(num_spokes=6)
 
+def test_to_dict_from_dict(example_protocol: Protocol):
+    proto_dict = example_protocol.to_dict()
+    new_protocol = Protocol.from_dict(proto_dict)
+    assert new_protocol == example_protocol
+
 @pytest.mark.parametrize("compact_representation", [True, False])
 def test_serialize_deserialize_protocol(example_protocol : Protocol, compact_representation: bool):
     assert example_protocol.from_json(example_protocol.to_json(compact_representation)) == example_protocol
--- a/tests/test_solution_analysis.py
+++ b/tests/test_solution_analysis.py
@@ -1,0 +1,80 @@
+from __future__ import annotations
+
+import pytest
+
+from openlifu.plan.param_constraint import ParameterConstraint
+from openlifu.plan.solution_analysis import SolutionAnalysis, SolutionAnalysisOptions
+
+# ---- Tests for SolutionAnalysis ----
+
+@pytest.fixture()
+def example_solution_analysis() -> SolutionAnalysis:
+    return SolutionAnalysis(
+        mainlobe_pnp_MPa=[1.1, 1.2],
+        mainlobe_isppa_Wcm2=[10.0, 12.0],
+        mainlobe_ispta_mWcm2=[500.0, 520.0],
+        beamwidth_lat_3dB_mm=[1.5, 1.6],
+        beamwidth_ele_3dB_mm=[2.0, 2.1],
+        beamwidth_ax_3dB_mm=[3.0, 3.1],
+        beamwidth_lat_6dB_mm=[1.8, 1.9],
+        beamwidth_ele_6dB_mm=[2.5, 2.6],
+        beamwidth_ax_6dB_mm=[3.5, 3.6],
+        sidelobe_pnp_MPa=[0.5, 0.6],
+        sidelobe_isppa_Wcm2=[5.0, 5.5],
+        global_pnp_MPa=[1.3],
+        global_isppa_Wcm2=[13.0],
+        p0_MPa=[1.0, 1.1],
+        TIC=0.7,
+        power_W=25.0,
+        MI=1.2,
+        global_ispta_mWcm2=540.0,
+        param_constraints={
+            "global_pnp_MPa": ParameterConstraint(
+                operator="<=",
+                warning_value=1.4,
+                error_value=1.6
+            )
+        }
+    )
+
+def test_to_dict_from_dict_solution_analysis(example_solution_analysis: SolutionAnalysis):
+    sa_dict = example_solution_analysis.to_dict()
+    new_solution = SolutionAnalysis.from_dict(sa_dict)
+    assert new_solution == example_solution_analysis
+
+@pytest.mark.parametrize("compact", [True, False])
+def test_serialize_deserialize_solution_analysis(example_solution_analysis: SolutionAnalysis, compact: bool):
+    json_str = example_solution_analysis.to_json(compact)
+    deserialized = SolutionAnalysis.from_json(json_str)
+    assert deserialized == example_solution_analysis
+
+
+# ---- Tests for SolutionAnalysisOptions ----
+
+
+@pytest.fixture()
+def example_solution_analysis_options() -> SolutionAnalysisOptions:
+    return SolutionAnalysisOptions(
+        standoff_sound_speed=1480.0,
+        standoff_density=990.0,
+        ref_sound_speed=1540.0,
+        ref_density=1020.0,
+        mainlobe_aspect_ratio=(1.0, 1.0, 4.0),
+        mainlobe_radius=2.0e-3,
+        beamwidth_radius=4.0e-3,
+        sidelobe_radius=2.5e-3,
+        sidelobe_zmin=0.5e-3,
+        distance_units="mm",
+        param_constraints={
+            "mainlobe_radius": ParameterConstraint(
+                operator=">=",
+                warning_value=1.5e-3,
+                error_value=1.0e-3
+            )
+        }
+    )
+
+def test_to_dict_from_dict_solution_analysis_options(example_solution_analysis_options: SolutionAnalysisOptions):
+    options_dict = example_solution_analysis_options.to_dict()
+    new_options = SolutionAnalysisOptions.from_dict(options_dict)
+    assert new_options == example_solution_analysis_options
--- a/db_dvc.dvc
+++ b/db_dvc.dvc
@@ -1,6 +1,6 @@
 outs:
-- md5: b5d5ad14d93e38f9600fea910603e6d6.dir
+- md5: bf109e310734c61c0eecff737562b425.dir
   nfiles: 257
   hash: md5
   path: db_dvc
-  size: 933880436
+  size: 933880511
--- a/src/openlifu/plan/param_constraint.py
+++ b/src/openlifu/plan/param_constraint.py
@@ -8,7 +8,7 @@ from openlifu.util.dict_conversion import DictMixin
 
 PARAM_STATUS_SYMBOLS = {
     "ok": "✅",
-    "warning": "⚠️",
+    "warning": "❗",
     "error": "❌"
 }
 
--- a/src/openlifu/plan/param_constraint.py
+++ b/src/openlifu/plan/param_constraint.py
@@ -26,12 +26,12 @@ class ParameterConstraint(DictMixin):
     def __post_init__(self):
         if self.warning_value is None and self.error_value is None:
             raise ValueError("At least one of warning_value or error_value must be set")
-        if self.operator in ['within','inside','outside','outside_inclusive']:
-            if self.warning_value and (not isinstance(self.warning_value, tuple) or len(self.warning_value) != 2 or (self.warning_value[0] >= self.warning_value[1])):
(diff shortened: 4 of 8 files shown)


ebrahimebrahim:
(If you get a chance it'd be nice to add a unit test for this also, since you probably already have some problematic code in hand)

sadhana-r (commented):
Looks good. I agree that it would be helpful to include a unit test to ensure that all of the same class attributes are defined before and after serialization.
Let me know if you're not able to address it now, then I'll approve and we can open a separate issue.

arhowe00:
> Looks good. I agree that it would be helpful to include a unit test to ensure that all of the same class attributes are defined before and after serialization. Let me know if you're not able to address it now, then I'll approve and we can open a separate issue.

The reason this wasn't caught by tests is because the following test, which _does_ have coverage for `to_dict`, passes when `allowed_roles` is empty. In fact, this test would pass with any values that are set to the default initialized values in the protocol (and passing would happen for any test object that has the same construction).

```python
@pytest.mark.parametrize("compact_representation", [True, False])
def test_serialize_deserialize_protocol(example_protocol : Protocol, compact_representation: bool):
    assert example_protocol.from_json(example_protocol.to_json(compact_representation)) == example_protocol
```

In particular:

- example_protocol.allowed_roles == [] (default)
- to_dict() omits allowed_roles
- from_json() sees no allowed_roles key → uses default ([])

So there are a few ways of fixing this. I'll changed the example protocol to include a more interesting `allowed_roles`, which led to a failing test that is fixed by modifying `Protocol.to_dict()`. I'm noting this because we **did** have coverage for this code, but the test passed anyway.

I also think I addressed #296 but I need to test and I'll re-request review when I check it out.

arhowe00:
Okay, I added some new changes and test cases for both the `allowed_roles` serialization fix in `Protocol` as well as resolved issue #296 which came up in the review for [SlicerOpenLIFU#284](https://github.com/OpenwaterHealth/SlicerOpenLIFU/pull/284). I wrote a new PR description at the top, so please reference before deciding whether to approve for merge.

tests/test_protocol.py
```diff
@@ -31,6 +31,11 @@ def example_session() -> Session:
 def example_wheel_pattern() -> Wheel:
     return Wheel(num_spokes=6)
 
+def test_to_dict_from_dict(example_protocol: Protocol):
+    proto_dict = example_protocol.to_dict()
+    new_protocol = Protocol.from_dict(proto_dict)
+    assert new_protocol == example_protocol
+
```
sadhana-r, line 38: I don't think this separate test is actually required now since `to_json` and `from_json` call to/from_dict. As you said, the main issue was that the example protocol wasn't catching the error.

tests/test_solution_analysis.py
```diff
@@ -0,0 +1,80 @@
...
+                warning_value=1.4,
+                error_value=1.6
+            )
+        }
+    )
+
+def test_to_dict_from_dict_solution_analysis(example_solution_analysis: SolutionAnalysis):
+    sa_dict = example_solution_analysis.to_dict()
+    new_solution = SolutionAnalysis.from_dict(sa_dict)
+    assert new_solution == example_solution_analysis
```
sadhana-r, line 43: Same here. We can remove this additional test since its covered by the serialze test.

sadhana-r (changes requested):
Looks good! Adding a minor comment.

arhowe00:
I added 2 commits on top that close #298 , an issue with the definition of `__post_init__` in the ParameterConstraint class definition. It also adds tests.

tests/test_protocol.py
arhowe00: This test actually *uniquely* fails when I introduced it because it exposes `SolutionAnalysisOptions` as a non-iterable.  The tests are not redundant because when I introduced this test, I got the unique failure:

```
__________________________________________________________________________________________________________________ test_to_dict_from_dict ___________________________________________________________________________________________________________________

example_protocol = Protocol(id='example_protocol', name='Example protocol', description='Example protocol created 30-Jan-2024 09:16:02', ...60, 66), yaw_step=2, planefit_dyaw_extent=14, planefit_dyaw_step=2, planefit_dpitch_extent=16, planefit_dpitch_step=7))

    def test_to_dict_from_dict(example_protocol: Protocol):
        proto_dict = example_protocol.to_dict()
>       new_protocol = Protocol.from_dict(proto_dict)

example_protocol = Protocol(id='example_protocol', name='Example protocol', description='Example protocol created 30-Jan-2024 09:16:02', ...60, 66), yaw_step=2, planefit_dyaw_extent=14, planefit_dyaw_step=2, planefit_dpitch_extent=16, planefit_dpitch_step=7))
proto_dict = {'allowed_roles': ['operator'], 'analysis_options': SolutionAnalysisOptions(standoff_sound_speed=1500.0, standoff_dens...01, distance_units='m', param_constraints={}), 'apod_method': Uniform(value=1.0), 'delay_method': Direct(c0=1540), ...}

tests/test_protocol.py:36:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

d = {'allowed_roles': ['operator'], 'analysis_options': SolutionAnalysisOptions(standoff_sound_speed=1500.0, standoff_dens...01, distance_units='m', param_constraints={}), 'apod_method': Uniform(value=1.0), 'delay_method': Direct(c0=1540), ...}

    @staticmethod
    def from_dict(d : Dict[str,Any]) -> Protocol:
        d["pulse"] = bf.Pulse.from_dict(d.get("pulse", {}))
        d["sequence"] = bf.Sequence.from_dict(d.get("sequence", {}))
        d["focal_pattern"] = bf.FocalPattern.from_dict(d.get("focal_pattern", {}))
        d["sim_setup"] = sim.SimSetup.from_dict(d.get("sim_setup", {}))
        d["delay_method"] = bf.DelayMethod.from_dict(d.get("delay_method", {}))
        d["apod_method"] = bf.ApodizationMethod.from_dict(d.get("apod_method", {}))
        seg_method_dict = d.get("seg_method", {})
        if "materials" in d:
            seg_method_dict["materials"] = seg.Material.from_dict(d.pop("materials"))
        d["seg_method"] = seg.SegmentationMethod.from_dict(seg_method_dict)
        d['param_constraints'] = {key: ParameterConstraint.from_dict(val) for key, val in d.get("param_constraints", {}).items()}
        if "target_constraints" in d:
            d['target_constraints'] = [TargetConstraints.from_dict(d_tc) for d_tc in d.get("target_constraints", {})]
        if "virtual_fit_options" in d:
            d['virtual_fit_options'] = VirtualFitOptions.from_dict(d['virtual_fit_options'])
        if "analysis_options" in d:
>           if "mainlobe_aspect_ratio" in d["analysis_options"]:
E           TypeError: argument of type 'SolutionAnalysisOptions' is not iterable

d          = {'allowed_roles': ['operator'], 'analysis_options': SolutionAnalysisOptions(standoff_sound_speed=1500.0, standoff_dens...01, distance_units='m', param_constraints={}), 'apod_method': Uniform(value=1.0), 'delay_method': Direct(c0=1540), ...}
seg_method_dict = {'class': 'Water', 'materials': {'air': {'attenuation': 7.5, 'density': 1.25, 'id': 'air', 'name': 'air', ...}, 'skull... 'tissue': {'attenuation': 0.3, 'density': 1050, 'id': 'tissue', 'name': 'tissue', ...}, ...}, 'ref_material': 'water'}

src/openlifu/plan/protocol.py:101: TypeError
================================================================================================================== short test summary info ==================================================================================================================
FAILED tests/test_package.py::test_version - AssertionError: assert '0.3.1.dev2+g3675aae' == '0.3.1.dev7+g...17a.d20250430'
FAILED tests/test_protocol.py::test_to_dict_from_dict - TypeError: argument of type 'SolutionAnalysisOptions' is not iterable
=============================================================================================================== 2 failed, 134 passed in 4.16s ===============================================================================================================
```

You can investigate this by checking out the commit the test was added and running `pytest`.  This probably could have also been solved by adding a test in to a different part of the codebase (probably in `SolutionAnalysisOptions`), but more coverage may be better because of this uniquely discovered failure. My guess for why it's failing while to/from_json don't is because `to_json` itself adds an extra layer of transformation through `json.dumps` which prevents the raw comparison of to/from_dict.

tests/test_solution_analysis.py
arhowe00: The addition of all the tests probably does solve the issue above, but for similar reasons as above there could exist a case in which the serialize test succeeds but the to/from_dict test fails.

sadhana-r (approved):
Parameter constraint changes look good to me!

PR #305, opened 2025-05-15
Added Version Retrieval to LIFU Interface

georgevigelette:
Highlights

added SDK version retrieval 
updated watertank script not to output prints every time it calls get temperature

### DIFF
--- a/src/openlifu/io/LIFUInterface.py
+++ b/src/openlifu/io/LIFUInterface.py
@@ -243,3 +243,12 @@ class LIFUInterface:
             self.txdevice.disconnect()
         if self.hvcontroller:
             self.hvcontroller.disconnect()
+
+    @staticmethod
+    def get_sdk_version() -> str:
+        """Get the SDK version from openlifu.__version__ or fallback."""
+        try:
+            import openlifu as m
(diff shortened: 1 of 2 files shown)


src/openlifu/io/LIFUInterface.py
```diff
@@ -243,3 +243,12 @@ def __exit__(self, exc_type, exc_value, traceback):
...
             self.hvcontroller.disconnect()
+
+    @staticmethod
+    def get_sdk_version() -> str:
+        """Get the SDK version from openlifu.__version__ or fallback."""
+        try:
+            import openlifu as m
+            return getattr(m, "__version__", "0.3.2")
+        except ImportError:
+            return "0.3.2"
```
peterhollender, line 254: Are these fallbacks necessary? Is there an expected case where openlifu doesn't import or doesn't have a `__version__`?

peterhollender (changes requested):
I think we can reduce this to just the minimum for `get_sdk_version` without the additional script updates and fallbacks.

ebrahimebrahim:
I am guessing this was meant to be overridden by #379

Issue #309, opened 2025-05-21
onnx runtime causing problems on windows


ebrahimebrahim:
From Seemal:

> Hi all, I've been using run_meshroom.py and I've been reaching an error with the onnxruntime dependency. It seems like it exists in pyproject.toml, and I've installed/uninstalled it multiple times individually to see if that would work. I also started from a fresh virtual environment and pip installed open-lifu python, but it still didn't work. The error doesn't specify what DLL is missing, I was wondering if anyone else has ran into this issue?

From Peter:

> the issue is with onnxruntime 1.22.0, which is the version grabbed by a naked pip install. We need to specify onnxruntime==1.18.0 until it gets fixed

See https://github.com/lshqqytiger/stable-diffusion-webui-amdgpu/issues/487
See https://github.com/microsoft/onnxruntime/issues/21270

PR #310, opened 2025-05-21
Pin onnxruntime version for windows DLL issue (#309)

ebrahimebrahim:
addresses #309

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -49,7 +49,7 @@ dependencies = [
   "OpenEXR",
   "scikit-image",
   "trimesh",
-  "onnxruntime",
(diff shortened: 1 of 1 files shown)


ebrahimebrahim:
When the onnx folks closed out that issue on their end, the solution they referenced was to install the correct version of the visual C++ 2019 runtime: https://onnxruntime.ai/docs/install/#requirements

So maybe our solution is either to insist on windows users having the latest version of visual C++ runtime, or pin onnxruntime to 1.18.0.

Peter's preference is to pin the version since that will be less of a burden on windows users

peterhollender (approved):

Issue #304, opened 2025-05-15
Enable reducing angular resolution in mesh reconstruction


ebrahimebrahim:
Expose an argument in `run_reconstruction` to take every n^th image, after sorting the images into a circle somehow (depends on how we address #303)

PR #312, opened 2025-05-23
Image file sorting and subsampling (#304)

dlichy:
This commit adds the `preprocess_image_paths` function, which sorts a list of image files either by numbers in the filename or by datetime metadata extracted from EXIF. It also supports downsampling the list to reduce angular resolution.


ebrahimebrahim:
heads up that I am about to rebase this (and reword a commit message to add an issue reference)

Issue #314, opened 2025-05-23
Return Python control after each Meshroom node


dlichy:
We should return Python control after each Meshroom node so that we can easily add callbacks, for example, to update a progress bar or better notify users of which Meshroom step failed.

Issue #313, opened 2025-05-23
Raise exception for invalid resolution combinations


dlichy:
For some combinations of input resolutions and depth map estimation downsampling configuations (combined depth resolution < 128) , Meshroom fails with an error.

```
Meshroom: aliceVision_depthMapEstimation: /opt/AliceVision_git/src/aliceVision/depthMap/cuda/imageProcessing/deviceMipmappedArray.cu:277: void aliceVision::depthMap::cuda_createMipmappedArrayFromImage(cudaMipmappedArray**, const aliceVision::depthMap::CudaDeviceMemoryPitched<aliceVision::depthMap::CudaRGBA, 2>&, unsigned int): Assertion `currentLevelArraySize.height == height' failed.
```

Instead of letting this happen, we should raise a Python exception when this situation is detected. That way, users can get a helpful message and fix their settings without digging through stack traces.

PR #315, opened 2025-05-23
Run Meshroom nodes sequentially, add timing, and restrict resize parameters (#313 and #314)

dlichy:
This PR returns Python control after each Meshroom node runs. It adds a `return_durations` argument to the `run_reconstruction` function, which, if set to True, returns a dictionary with the runtime of each node. Additionally, `run_reconstruction` now raises a `ValueError` if the combination of `input_resize_width` and selected pipeline would result in a Meshroom CUDA error.


ebrahimebrahim:
(heads up that I am about to rebase and force push)

PR #327, opened 2025-05-29
Fix #134: Prevent duplicate target IDs in Session

### LINKED ISSUE #134: It should not be posssible to duplicate target IDs listed in a session
When listing target IDs in a session, it should not be possible to write duplicate target IDs.

This could be enforced within the Session class somehow, or it could just be enforced at session write time.

HussainAther:
This pull request fixes issue [[#134](https://github.com/OpenwaterHealth/OpenLIFU-python/issues/134)](https://github.com/OpenwaterHealth/OpenLIFU-python/issues/134) by adding a check in the `Session` class to prevent duplicate target IDs.

### 🔧 Changes Made

* Added a validation step in `Session.__post_init__` to ensure all target IDs are unique.
* Added a test case in `test_session.py` to verify the behavior.
* Passed all pre-commit checks and linters (`pre-commit`, `ruff`, etc.).

### ✅ Why This Matters

Ensuring unique target IDs helps prevent data integrity issues and logic bugs during session processing.

Let me know if any changes or refactors are needed.

### DIFF
--- a/tests/test_session.py
+++ b/tests/test_session.py
@@ -1,0 +1,14 @@
+from __future__ import annotations
+
+import pytest
+
+from openlifu.db.session import Session
+from openlifu.geo import Point
+
+
+def test_duplicate_target_ids_raises():
+    target1 = Point(id="T1", position=[0, 0, 0])
+    target2 = Point(id="T1", position=[1, 1, 1])  # Duplicate ID
+
+    with pytest.raises(ValueError, match="Duplicate target IDs found"):
+        Session(targets=[target1, target2])


ebrahimebrahim:
Hi @HussainAther and thanks for your contribution!

The title of the PR is not right, and the author of the commit, as well as the commit message tes, seem to be not right. Perhaps a rebase gone wrong?

The test should go into `test_database.py` that already exists, since all our other tests of `Session` live there

The stated approach makes sense and test looks good. Now to add the code that makes it pass :slightly_smiling_face: 

Feel free to tag me as reviewer when ready

Issue #334, opened 2025-06-05
Add fields to solution analysis


peterhollender:
Solution Analysis would benefit from communicating information to the user about the focal position of the target (nominal and actual), the duty cycle of the sequence (implied in the protocol definition, but useful to see, particularly if the voltage comes back out of bounds), and the overall sequence duration.

PR #335, opened 2025-06-05
enhance_analysis for #334

peterhollender:
This adds the following attributes to `SolutionAnalysis`:

- `target_position_lat_mm`
- `target_position_ele_mm`
- `target_position_ax_mm`
- `focal_centroid_lat_mm`
- `focal_centroid_ele_mm`
- `focal_centroid_ax_mm`
- `duty_cycle_pulse_train_pct`
- `duty_cycle_sequence_pct`
- `sequence_duration_s`

Some of the display format for other dimension based parameters are updated as well.

Significant figures on all items in solution analysis when generating a table are also refined.

### DIFF
--- a/src/openlifu/plan/solution.py
+++ b/src/openlifu/plan/solution.py
@@ -17,6 +17,7 @@ from openlifu.plan.param_constraint import ParameterConstraint
 from openlifu.plan.solution_analysis import (
     SolutionAnalysis,
     SolutionAnalysisOptions,
+    find_centroid,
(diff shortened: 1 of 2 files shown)


ebrahimebrahim (approved):
Looks good, and it ran and displayed the new things in the SlicerOpenLIFU solution analysis table.

(This is not a "semantic" review -- I am not checking the physical correctness of the analysis but just the soundness of the code.)

PR #348, opened 2025-06-13
serialize TargetConstraints #347

### LINKED ISSUE #347: Protocols do not serialize TargetConstraints
when protocols contain TargetConstraints, they do not serialize correctly, because `protocol.to_dict()` doesn't turn the list of targetconstraints to dicts

peterhollender:
Closes #347

### DIFF
--- a/src/openlifu/plan/protocol.py
+++ b/src/openlifu/plan/protocol.py
@@ -114,7 +114,7 @@ class Protocol:
             "apod_method": self.apod_method.to_dict(),
             "seg_method": self.seg_method.to_dict(),
             "param_constraints": {id: pc.to_dict() for id, pc in self.param_constraints.items()},
-            "target_constraints": self.target_constraints,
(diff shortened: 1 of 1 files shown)


ebrahimebrahim (changes requested):
Thank you! This would have been caught by the test

https://github.com/OpenwaterHealth/OpenLIFU-python/blob/c83a1337118e78051dc79b6c68eaef3d5f781ac9/tests/test_protocol.py#L39-L41

but I am guessing it didn't get caught because the protocol used for testing doesn't have any target constraints:

https://github.com/OpenwaterHealth/OpenLIFU-python/blob/c83a1337118e78051dc79b6c68eaef3d5f781ac9/tests/resources/example_db/protocols/example_protocol/example_protocol.json#L91

Can you add some target constraints to `tests/resources/example_db/protocols/example_protocol/example_protocol.json` in this PR? Or I can, lmk

ebrahimebrahim (approved):
ty for updating the test data!

PR #355, opened 2025-06-29
feat: Add mainlobe-to-sidelobe pressure and intensity ratios

### LINKED ISSUE #345: Analysis of Ratios
When running beamforming analysis, we may want to know the ratio of the sidelobe to mainlobe pressure and/or intensity. We can already set thresholds for warnings on the sidelobes and mainlobe, but it would be convenient to set those as relative values.

peterhollender:
Adds the calculation and reporting of mainlobe-to-sidelobe pressure and intensity ratios to the SolutionAnalysis.

- New fields in SolutionAnalysis dataclass.
- PARAM_FORMATS updated for table display.
- Ratio calculation logic added to Solution.analyze(), handling division by zero.
- Tests updated and new tests added for ratio calculations.

Closes #345


peterhollender:
Yeah, that makes sense as intended behavior.

PR #361, opened 2025-07-21
Feature/protocol to table

### LINKED ISSUE #360: Protocol to_table
The `plan.solution_analysis.SolutionAnalysis` class has a nice `to_table` method that creates a Pandas Dataframe with nicely formatted display names and numbers.  In SlicerOpenLIFU, this table is useful for displaying output, as we can extract the important display columns and pipe them to the UI. We should do the same thing with `plan.protocol.Protocol`. Nice formatting of a Protocol for display will allow us to add convenient display of Protocol settings. There will need to be recursive retrieval of formatted tables from some attributes, as those attributes will be instances of classes with specific subclass-dependent properties. Some of the classes, like `Sequence` and `Pulse`, already have `get_table` methods, so this should be extended throughout the codebase as needed.

peterhollender:
Closes #360

### DIFF
--- a/src/openlifu/seg/seg_methods/uniform.py
+++ b/src/openlifu/seg/seg_methods/uniform.py
@@ -18,7 +18,7 @@ class UniformSegmentation(SegmentationMethod):
         :returns: Pandas DataFrame of the segmentation method parameters
         """
         records = [{"Name": "Type", "Value": "Uniform Tissue", "Unit": ""},
-                   {"Name": "Reference Material", "Value": self._ref_material, "Unit": ""}]
+                   {"Name": "Reference Material", "Value": self.ref_material, "Unit": ""}]
         return pd.DataFrame.from_records(records)
 
 class UniformTissue(UniformSegmentation):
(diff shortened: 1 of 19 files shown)


src/openlifu/seg/seg_methods/uniform.py
```diff
@@ -10,16 +11,44 @@ class UniformSegmentation(SegmentationMethod):
     def _segment(self, volume: xa.DataArray):
         return self._ref_segment(volume.coords)
 
+    def to_table(self) -> pd.DataFrame:
+        """
+        Get a table of the segmentation method parameters
+
+        :returns: Pandas DataFrame of the segmentation method parameters
+        """
+        records = [{"Name": "Type", "Value": "Uniform Tissue", "Unit": ""},
```
arhowe00, line 20: Why are we returning the reference material here but not for the `UniformTissue`/`UniformWater`? Also, are we sure that all `UniformSegmentation` are always "Uniform Tissue" Type?

arhowe00 (approved):
Approved for merging, but I left a question @peterhollender you should look at before doing so.

ebrahimebrahim (approved):
Looks good, thanks! Squashing and merging for this one

PR #363, opened 2025-07-22
feat: Remove special cases from SegmenationMethod.from_dict

### LINKED ISSUE #362: Remove special cases from `SegmenationMethod.from_dict`
`SementationMethod.from_dict` includes special-case handlers for specific subclasses:
```     
if short_classname in ["UniformWater", "UniformTissue"]:
    d.pop("ref_material")
```

This violates the spirit of `SegmentationMethod.from_dict`, as "UniformWater" and "UniformTissue" (and any new subclasses!) have to have part of their `from_dict` behavior implemented in the abstract `SegmentationMethod` classes. Each subclass should instead, if needed, properly specify it's `to_dict` method so that it only saves the necessary attributes for its own reconstruction. In the cases of `UniformWater` and `UniformTissue`,  `reference_material` should be excluded, as it is implicit in the class and not available as a constructor input.

peterhollender:
Removes the special-case handling for `UniformWater` and `UniformTissue` in `SegmentationMethod.from_dict`. Each subclass now overrides `to_dict` to exclude the `ref_material` attribute, which is implicit in the class and not a constructor argument.

`SegmentationMethod.from_dict` now checks for unexpected keywords and handles them based on the `on_keyword_mismatch` parameter. This allows for graceful handling of old JSON files that may contain the `ref_material` keyword.

Closes #362

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -94,8 +94,7 @@ addopts = ["-ra", "--showlocals", "--strict-markers", "--strict-config"]
 xfail_strict = true
 filterwarnings = [
   "error",
-  "ignore:Unsupported Windows version.*ONNX Runtime supports Windows 10 and above, only.:UserWarning",
-  "ignore:Ignoring unexpected keyword arguments for UniformWater.*:UserWarning",
+  "ignore:Unsupported Windows version.*ONNX Runtime supports Windows 10 and above, only.:UserWarning"
 ]
 log_cli_level = "INFO"
 testpaths = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -95,6 +95,7 @@ xfail_strict = true
 filterwarnings = [
   "error",
   "ignore:Unsupported Windows version.*ONNX Runtime supports Windows 10 and above, only.:UserWarning",
+  "ignore:Ignoring unexpected keyword arguments for UniformWater.*:UserWarning",
 ]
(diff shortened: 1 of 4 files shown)


arhowe00 (approved):
The test cases look good. All I did was rename commit titles. I'll approve this for merging, feel free to comment on below if anything raises a concern:

1. There are many lines of code designed specifically to handle how `ref_material` being set, when what this field expresses can be made simpler. What the inheritance hierarchy means is clashing with `ref_material` being settable in certain cases. We could remove the subclassing and ABC entirely, with everything being specified in a SegmentationMethod, or we could make this feature non-settable (even in constructors), make it class level for all specializations of `SegmentationMethod`.

2. I'm not sure whether the

> graceful handling of old JSON files that may contain the `ref_material` keyword.

is an approach we should take for all things going forward, as it adds a lot of debt if we add handling to old versions of data. I think the project is small enough right now that we can keep the debt low and remove the field from the existing (small number) of JSON files. I'll approve it for merging anyway as this is cautionary but not a problem.

ebrahimebrahim (changes requested):

ebrahimebrahim:
Sounds good! You can re-request review from me when done with additional changes

(we won't do any force-pushing till then -- the branch is in your hands, Peter. @arhowe00)

ebrahimebrahim (approved):

PR #365, opened 2025-07-25
Remove hard-coded use of "lat", "ele", "ax"

### LINKED ISSUE #364: Inconsistent use of dimension names
When creating the local transducer dimensions, objects like `Point`s default to `x`,`y`,`z`, but other methods use `lat`, `ele`, `ax`. We may be able to switch everything to xyz and make it cleaner throughout.

peterhollender:
Updates the codebase to be less expectant of particular dimension names, and removes the ability to specify them in SimSetup, as this level of configuration doesn't provide enough benefit to overcome the risk of confusion
Closes #364


ebrahimebrahim (approved):
Looks great, thank you! 

(force pushing to adjust a commit message)

PR #371, opened 2025-08-11
Fix test scripts to call start_sonication() which now includes HV turn on command.

alkagan:
Changed all start_trigger() calls to start_sonication() calls in test scripts, added function parameter to start_sonication() to select using external power supply. (#75)


ebrahimebrahim (commented):
LGTM; leaving final approval to Peter

alkagan:
> (the commit message subject line is very long though. not too important but generally best to keep the subject (the first line) very short and add details if needed in the message body)

Modified commit message to be neater

Issue #367, opened 2025-07-30
Virtual Fit Slow


peterhollender:
Virtual fit takes a few minutes to run, and it is unclear why the runtime is so long. I'm guessing that the estimation of normals to curved skin points is the bottleneck, but it feels like there is room for improvement. We should profile performance and see if we can't improve.

ebrahimebrahim:
Agreed there is room for improvement -- I never had to deal with a few _minutes_ though. It's usually under a minute for me. But it depends on the size of the search space and the resolution of the spherical interpolator.

The normals estimation is not a bottleneck I think. That's all vectorized pretty well.

Is it hanging mainly on "building spherical interpolator" or on "searching through candidates"?

peterhollender:
> The pose search taking longer makes sense -- the iteration over candidate poses is not vectorized.
> 
> But from your options that's not tooo dense of a search grid, it looks like ~100 pts, so I do worry that maybe something is wrong. Worth investigating.
> 
> If nothing is wrong a low hanging fruit is to parallelize the search through poses, since each poses is really analyzed independently.

Please investigate. In the MATLAB code, I was able to build a tool with a little navigation GUI that would update the position of the array along the surface of the head in basically real-time, including updating a fancier virtual fitting that did element-wise raytracing to determine if an element path was blocked by the sinuses. The computation time per search position is way higher than expected.

ebrahimebrahim:
I've got two PRs up for this now, but I need to know more to see if they will solve the problem:

- https://github.com/OpenwaterHealth/OpenLIFU-python/pull/374 (big speedup for intel cpu, using raytracing)
- https://github.com/OpenwaterHealth/SlicerOpenLIFU/pull/474 (small speedup to pose search while in slicer)

As you can see in the profiling info at https://github.com/OpenwaterHealth/OpenLIFU-python/pull/374, under the same conditions as yours in this issue, the pose search takes a few seconds, not _minutes_, for me, even before these changes. So I suspect that something else is going on, possibly something to do with linear algebra libraries on each of our systems.

[...]

PR #374, opened 2025-08-19
Speed up virtual fitting

ebrahimebrahim:
This is work towards #367 

The main change is to introduce embree CPU raytracing via embreex which is integrated with trimesh. Trimesh is already a dependency of openlifu so this came almost for free.

The raytracing speeds up both construction and use of the spherical interpolator. Here is some profiling:

#### previous timings

Conditions for all these:
- Running virtual fit from SlicerOpenLIFU my ubuntu 22.04 machine. 
- I am using the example currently (as of 94209171bbc47bb18954580fb153d3f0f930c54f) in the dvc data, specifically the session `neuromod_2x_demo`. This contains the `VirtualFitOptions` that were slow for Peter in #367

|task|time|
|-|-|
|skin interpolator construction| 15s |
|pose search | 7s |

(About 8s of the 15s in skin interpolator construction are not spent tracing rays but rather just building the scipy LinearNDInterpolator, which uses the points to construct a triangulation.)

#### after just the thread cap introduced in https://github.com/OpenwaterHealth/SlicerOpenLIFU/pull/474

|task|time|
|-|-|
|skin interpolator construction| 15s |
|pose search | **4s** |

It just shaved off a few seconds from the pose search.

#### after also including this PR

|task|time|
|-|-|
|both| **1.5s** |

(and most of that 1.5s seems to be spent on Slicer updating model transforms, so the virtual fit algorithm part is quite fast. see gif here: https://github.com/OpenwaterHealth/SlicerOpenLIFU/pull/474#issuecomment-3200701780)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -49,7 +49,7 @@ dependencies = [
   "OpenEXR",
   "scikit-image",
   "trimesh",
-  "embreex; platform_machine=='x86_64'",
+  "embreex; platform_machine=='x86_64' or platform_machine=='AMD64'",
   "onnxruntime==1.18.0",
 ]
 


ebrahimebrahim:
@sadhana-r This is most easily reviewed by reviewing the SlicerOpenLIFU companion PR: https://github.com/OpenwaterHealth/SlicerOpenLIFU/pull/474

A cursory code review is sufficient, along with a test running this new virtual fit from SlicerOpenLIFU via the companion PR. I will add a couple of comments to provide some context

src/openlifu/seg/skinseg.py
```diff
@@ -18,13 +18,14 @@
 import numpy as np
 import skimage.filters
 import skimage.measure
+import trimesh
```
ebrahimebrahim, line 21: trimesh is not a new dependency here -- we were already using it elsewhere

src/openlifu/virtual_fit.py
```diff
@@ -329,17 +329,19 @@ def progress_callback(progress_percent : int, step_description : str): # noqa: A
     points_asl = np.zeros((num_search_points,3), dtype=float) # search grid points in ASL coordinates
     normals_asl = np.zeros((num_search_points,3), dtype=float) # normal vector of the plane that is fitted at each point, in ASL coordinates
 
+    # Pre-build plane fitting grid that will be used repeatedly in the search loop below.
+    # This grid lives in the spherical coordinate basis theta-phi plane at each point.
+    # Below we will project it onto the skin surface for each search point.
+    dtheta_sequence = np.arange(-planefit_dyaw_extent, planefit_dyaw_extent + planefit_dyaw_step, planefit_dyaw_step)
+    dphi_sequence = np.arange(-planefit_dpitch_extent, planefit_dpitch_extent + planefit_dpitch_step, planefit_dpitch_step)
+    dtheta_grid, dphi_grid = np.meshgrid(dtheta_sequence, dphi_sequence, indexing='ij')
```
ebrahimebrahim, line 337: the construction of this point grid was previously inside the `for i in range(num_search_points):` just below, and I moved it out because it is the same grid each time.

src/openlifu/seg/skinseg.py
```diff
@@ -420,3 +438,45 @@ def spherical_interpolator_from_mesh(
     )
 
     return interpolator
+
+def _spherical_interpolator_from_mesh_embree(surface_mesh : vtk.vtkPolyData) -> Callable:
```
ebrahimebrahim, line 442: this function's implementation doesn't need to be reviewed closely because the unit test `test_spherical_interpolator_from_mesh` covers it well; we now run that test on both `_spherical_interpolator_from_mesh_cell_locator` (the old method) and `_spherical_interpolator_from_mesh_embree` (the new method)

ebrahimebrahim:
Oh so you do reproduce Peter's problem -- that's perfect!

sadhana-r:
Oh right, its amd64. This works!

sadhana-r (approved):

PR #386, opened 2025-09-19
377 support local versions of the resources that are downloaded

ebrahimebrahim:
(description incoming)


sadhana-r (approved):
Tested this via https://github.com/OpenwaterHealth/SlicerOpenLIFU/pull/479 and it worked as expected

PR #393, opened 2025-10-24
Feature/multiple module enhancements

alkagan:
Add UART command to query number of devices connected


src/openlifu/io/LIFUUart.py
```diff
@@ -293,6 +294,34 @@ def list_vcp_with_vid_pid(self):
...
+                        self.connect()
+                        log.info("Reconnected to UART on %s after reset.", self.port)
+                        return True
+                    except Exception as e:
+                        log.warning("Reconnect attempt failed: %s", e)
+            log.error("Failed to reconnect UART after reset.")
+            return False
+        except Exception as e:
+            log.error("reopen_after_reset() error: %s", e)
+            return False
```
ebrahimebrahim: Better not to [swallow up arbitrary exceptions like this](https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/broad-exception-caught.html), because it can unintentionally hide future bugs and make debugging difficult.

Best is to catch the specific exception you are trying to catch

An also ok approach would be to re-raise the exception here instead of `return False`. (but then what's the point of these lines? the python traceback will already show that the exception occured in `reopen_after_reset()` so there is no need to log that fact manually)

(Same goes for the other line above that swallows arbitrary exceptions)

ebrahimebrahim (changes requested):
Leaving review of notebooks and hardware interface code to others; it looks good from my perspective of SlicerOpenLIFU

Don't worry about commit message formatting because I can squash-n-merge and fix it at the end.

Issue #274, opened 2025-04-08
Create database validity checker


ebrahimebrahim:
This will support https://github.com/OpenwaterHealth/SlicerOpenLIFU/issues/235

Issue #338, opened 2025-06-06
Documentation (SOFTREQ-134)


peterhollender:
Software shall be sufficiently documented

Notes:
nan

Updated on 2025-06-06 16:27:00

PR #395, opened 2025-10-28
Update README.rst

stahir715:
Changes to description about using DVC and how downloading the sample database is only intended to be used by developers. #274

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -126,19 +126,17 @@ Version control of database using DVC (Data Version Control)
 
 Data Version Control (DVC) is a data management tool that is meant to be run alongside Git.
 In this project, DVC is used to link changes in the code to specific versions of a sample database containing example project files.
-DVC can be used when this project is installed in Dev mode. You can read more about DVC and how to use it `here <https://dvc.org/doc/start>`_.
-**Note:** Remote access to the sample database stored on google drive is currently restricted. Access requires a :code:`gdrive_client_secret`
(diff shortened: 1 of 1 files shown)


stahir715:
> Thanks for this update!
> 
> I'd suggest instead of saying "dev mode" to say that DVC is intended to be used by developers. Saying "dev mode" makes it sound as though there's a separate mode in which the library can be installed
> 
> Also [we require](https://github.com/OpenwaterHealth/OpenLIFU-python/blob/main/.github/CONTRIBUTING.md#commit-and-pull-request-expectations) that commit messages mention a software requirement issue, so that all code changes can ultimately be linked to the matrix SRS. So this commit for example could mention #338
> 
> LMK if you want me to make the edits or if you want to update the PR yourself. I can also fix the commit message or do so while merging at the end

Thanks for the feedback! Do i need to format the software requirement issue in a certain way, or as long as it mentions it in any form it works for the matrix SRS? I'll go ahead and update it myself :)

ebrahimebrahim:
> Thanks for the feedback! Do i need to format the software requirement issue in a certain way, or as long as it mentions it in any form it works for the matrix SRS? I'll go ahead and update it myself :)

ty!

Just writing #338 alone is enough, whether it's in the commit message title or body either is fine

You can see in the commit history how it was done: https://github.com/OpenwaterHealth/OpenLIFU-python/commits/main/

PR #397, opened 2025-10-30
Relax temperature increase cutoffs for tx and console. (#75)

alkagan:
This script is being used for testing - I relaxed the temperature increase cutoffs since they were too restrictive.


ebrahimebrahim (approved):
Looks great! Will leave a couple of days for Peter to comment before merging

PR #398, opened 2025-11-10
Add DICOM support for volume additions

### LINKED ISSUE #396: Enable adding volume from DICOM
> Hey, when we do "Add Volume", the file browser has a dropdown that includes "DICOM" as a valid format, but the window only lets you select a single file. Does that DICOM loader work? Is it expected to?

It doesn't currently work, but we should make it work.

OpenLIFU-python might only support having a single file represent a volume, but that might be just fine because once the user adds a volume from DICOM we can always write to the openlifu database as a nifti.

arhowe00:
Closes #396 

Dicom series and dicom files are converted to nifti format and stored in the database. 

## Data

The added test data comes from the data used for pydicom's unit testing:

### CT_small.dcm:
- https://github.com/pydicom/pydicom/raw/main/src/pydicom/data/test_files/CT_small.dcm

### dicom_series files (6293 and 6924):
- https://github.com/pydicom/pydicom/raw/main/src/pydicom/data/test_files/dicomdirtests/98892001/CT2N/6293
- https://github.com/pydicom/pydicom/raw/main/src/pydicom/data/test_files/dicomdirtests/98892001/CT2N/6924

### DIFF
--- a/src/openlifu/util/volume_conversion.py
+++ b/src/openlifu/util/volume_conversion.py
@@ -46,13 +46,16 @@ def is_dicom_file_or_directory(path: PathLike) -> bool:
     return False
 
 
-def extract_affine_from_dicom(dicom_slices: list) -> np.ndarray:
+def extract_affine_from_dicom(
+    dicom_slices: list[tuple[int, np.ndarray, pydicom.Dataset]]
+) -> np.ndarray:
     """
     Extract the affine transformation matrix from DICOM header information.
     Converts from DICOM LPS (Left-Posterior-Superior) to NIfTI RAS.
 
     Args:
-        dicom_slices: List of tuples (instance_number, pixel_array, dicom_dataset)
+        dicom_slices: List of tuples (instance_number, pixel_array, header) where
+            header is a pydicom.Dataset containing DICOM metadata tags
 
     Returns:
         4x4 affine transformation matrix mapping voxel coordinates to RAS world coordinates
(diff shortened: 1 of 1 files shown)


src/openlifu/db/database.py
```diff
@@ -24,6 +28,95 @@
...
+                        f.seek(128)
+                        if f.read(4) == b'DICM':
+                            return True
+                except OSError:
+                    continue
+
+    return False
+
+
+def convert_dicom_to_nifti(input_path: PathLike, output_filepath: PathLike) -> None:
```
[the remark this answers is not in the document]
arhowe00: Great idea, done

ebrahimebrahim (changes requested):
Thanks for the updates

I tried testing this in SlicerOpenLIFU itself, and it looks like something is off with the affine:

image

Here's how I got that screenshot:

- Used Slicer to convert the nifti volume in our dvc data to DICOM. [Here is the resulting dicom](https://drive.google.com/file/d/1ZBzFp0nhmOgPHksE_dXtiwrHFiUcs0A2/view?usp=sharing) (shared with @arhowe00)
- Made a [tweak](https://github.com/OpenwaterHealth/SlicerOpenLIFU/commit/b880246f61737b6300e05f9f0c7b91b674107ed4) to SlicerOpenLIFU to enable specifying a folder rather than a file when adding volume. (Later we will have to come up with a way to allow both)
- Used the _add volume_ dialog in SlicerOpenLIFU to add a volume. This internally calls `Database.write_volume` so I could have just done that instead of using SlicerOpenLIFU.
- Load the volume (which has now been converted to NIFTI) back into Slicer

The right affine metadata is definitely there in the dicom file, because when I used Slicer's dicom importer it came out matching the original nifti.

So something must be wrong with the affine handling in the dicom-to-nifti conversion in this PR

src/openlifu/util/volume_conversion.py
```diff
@@ -0,0 +1,157 @@
...
+                        f.seek(128)
+                        if f.read(4) == b'DICM':
+                            return True
+                except OSError:
+                    continue
+
+    return False
+
+
+def extract_affine_from_dicom(dicom_slices: list) -> np.ndarray:
```
ebrahimebrahim: Is it possible to complete this type annotation beyond just `list`? Ignore this comment if it's too inconvenient

src/openlifu/util/volume_conversion.py
```diff
@@ -0,0 +1,157 @@
...
+    """
+    Extract the affine transformation matrix from DICOM header information.
+    Converts from DICOM LPS (Left-Posterior-Superior) to NIfTI RAS.
+
+    Args:
+        dicom_slices: List of tuples (instance_number, pixel_array, dicom_dataset)
```
ebrahimebrahim: Can you elaborate a little on what `dicom_dataset` means here? Is it a metadata format from pydicom?

PR #404, opened 2025-12-03
Feature/vmonitor

georgevigelette:
Added ADC monitoring for HIGH voltage rails.  
Added calibrated power supply capabilities.


ebrahimebrahim:
Yes we prioritized it and as you can see it had been merged already. The next release is on its way.

PR #441, opened 2026-04-01
P/kwave0p6

### LINKED ISSUE #438: Upgrade kwave to v0.6.x
>  k-wave-python can now be run completely in Python and no longer depends on the k-wave binary executables.

The numpy<2 pin won't do any longer, and this will also cause us to unpin onnxruntime. See
- https://github.com/OpenwaterHealth/openlifu-python/commit/8b21331b70a575f7b89617e8884052b56e49288c
- https://github.com/OpenwaterHealth/openlifu-python/issues/309

(Also, see Peter's existing branch on upgrading kwave before getting started with this)

peterhollender:
Closes #438 
Requires testing of onnxruntime dependencies/numpy version, and testing in Slicer before merging

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -110,6 +110,7 @@ celerybeat.pid
 .env
 .venv
 env/
+env*
 venv/
 ENV/
 env.bak/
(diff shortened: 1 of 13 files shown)


ebrahimebrahim:
@peterhollender Is this PR stale or should we be reviving it?

Issue #440, opened 2026-03-31
Load txm definition from device


peterhollender:
With the upgrades in the latest txm firmware, we can store calibration data on the devices themselves. We should implement ways to read and write this data to remove dependence on database storage of transducer definitions.

PR #443, opened 2026-04-07
P/frequency dependent sensitivity

### LINKED ISSUE #439: Add frequency-dependent sensitivity
Elements and Transducers technically have frequency-dependent sensitivity through the inclusion of an impulse response, but since we always send single-frequency pulses, it will be much more compact and precise to represent frequency-dependent sensitivity directly. This will improve representation of transducer performance, at both the nominal frequency and the peak frequency, which may differ.

peterhollender:
closes #439 #440


peterhollender:
The rebasing is throwing me for a loop. Could I get some help with the commit cleanup?

ebrahimebrahim:
Sure thing, heads up that I will force-push to this branch shortly.

ebrahimebrahim:
@peterhollender Want me to review this yet? If so, which makes sense to review and merge first, this one or #444 ?

(Before the next slicer extension release, I am thinking we should get in all of these big openlifu-python changes, so that we can test/upgrade things in the extension all at once.)

src/openlifu/xdc/element.py
```diff
@@ -2,14 +2,40 @@
...
+    if isinstance(sensitivity, list):
+        freqs, values = zip(*sensitivity)
+        freqs = np.array(freqs, dtype=np.float64)
+        values = np.array(values, dtype=np.float64)
+        if frequency in freqs:
+            idx = np.where(freqs == frequency)[0][0]
+            return float(values[idx])
+        else:
+            return float(np.interp(frequency, freqs, values, left=values[0], right=values[-1]))
+    return float(sensitivity)
```
ebrahimebrahim, line 18: This can be simplified a lot, because `np.interp` will do the right thing

```suggestion
def sensitivity_at_frequency(sensitivity: float | List[tuple[float, float]], frequency: float) -> float:
    if isinstance(sensitivity, list):
        freqs, values = zip(*sensitivity)
        return float(np.interp(frequency, freqs, values))
    return float(sensitivity)
```

(also the floating point equality check `freqs == frequency` can be fragile -- but anyway it doesn't matter because the interp handles things correctly)

(Note that np.inter requires `freqs` to be monotonically increasing. If that's not guaranteed then you'd need to sort it at the top of this function.)

src/openlifu/sim/kwave_if.py
```diff
@@ -173,9 +176,10 @@ def run_simulation(arr: xdc.Transducer,
...
-                    crosstalk_arr.elements += [arr.elements[dst_idx].copy()]
-                    crosstalk_mat = np.vstack((crosstalk_mat, arr.crosstalk_frac*source_mat[src_idx,:]))
-        arr = crosstalk_arr
+         #           crosstalk_arr.elements += [arr.elements[dst_idx].copy()]
+                    crosstalk_mat[dst_idx,:] += arr.crosstalk_frac*source_mat[src_idx,:]
+                    #crosstalk_mat = np.vstack((crosstalk_mat, arr.crosstalk_frac*source_mat[src_idx,:]))
+        #arr = crosstalk_arr
         source_mat = crosstalk_mat
     karray = get_karray(arr,
                         translation=array_offset,
```
ebrahimebrahim: clean up and remove commented out code? committing to the new way of computing crosstalk

src/openlifu/xdc/util.py
```diff
@@ -1,11 +1,243 @@
...
 from openlifu.xdc.transducer import Transducer
 from openlifu.xdc.transducerarray import TransducerArray
 
+FOCAL_GAIN_LUT = xa.DataArray.from_dict(
+    {'dims': ('f0', 'crosstalk'),
+ 'attrs': {},
+ 'data': [[2.807589054107666,
+   3.2286391258239746,
+   3.649686813354492,
+   3.8181092739105225,
```
ebrahimebrahim, line 19: Totally optional but consider saving the lookup table into a data file (json, csv, etc.) to make the source code file shorter.

If you chose to do that:
- you would read the file here using `importlib.resources` to locate it
- you would ensure it's bundled int he install by adding it to the `[tool.setuptools.package-data]` section. in case hatchling doesn't bnundle it automatically

ebrahimebrahim (changes requested):
This is great!
A couple of minor points and this should be good to go.

src/openlifu/xdc/util.py
peterhollender: That's a good idea, but I don't want to rock the boat. Opening #448

ebrahimebrahim (approved):
Looks great! Opening #448 for later is fine. 

One thing to still be super careful about and that you might want to address: `np.interp` _assumes_ the `freqs` are sorted monotonically. Sorting is not currently enforced anywhere, e.g. you could do it in `Element.__post_init__`.

ebrahimebrahim:
(adding the sorting befroe merging actually)

Issue #10, opened 2024-06-17
package dependencies are not minimal


ebrahimebrahim:
The package dependencies that we list are not minimal. Things are included that are not really needed to install and use the `openlifu` package itself but only needed for other purposes such as running example notebooks or generating documentation. 

https://github.com/OpenwaterHealth/OpenLIFU-python/blob/46d73e8d70886911890a8e20351ea15d27ed8460/pyproject.toml#L5-L18

- We should pare down this list to only the essentials.
- Dependencies needed for running example notebooks can be put somewhere else like a `requirements.txt` inside the notebook directory, or even just notebook code cells with a `!pip install` in them.
- Dependencies needed for generating documentation, running tests, running linting, etc. can be put into their own groups of [optional dependencies](https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#optional-dependencies). This part will be done as part of the resolution of #5 and doesn't need to be considered part of this issue.

ebrahimebrahim:
Closed by #444

PR #444, opened 2026-04-07
P/init install cleanup

peterhollender:
This is a fairly major change to the project, designed to address issues #8 and #10 

- Trims unnecessary parts of the project, including
  - legacy scripts
  - the io module (now covered in `openlifu-sdk`)
  - unused parts of Database (gridweights)
- Reorganizes `virtual_fit` module into `seg`
- Eliminates "eager" imports in `openlifu.__init__`, to prevent the many-seconds-long importing when accessing any submodule.
- Moves importing of heavier-weight packages (which are being made optional by this branch) into the methods that call them, so that the objects can be partially-used without having those dependencies imported.
- Introduces sets of optional-dependencies to `pyproject.toml` that allow for reduced dependency installation. The `openlifu-test-app` project, for example, can package only the hardware-communication required dependencies by using the `[io]` option. The base application as used in Slicer should now be installed with the `[app]` option.

Code that touches `openlifu.virtual_fit` needs to be refactored to use `openlifu.seg.virtualfit`, and code that used things like `openlifu.Transducer` will need to be changed to use `openlifu.xdc.Transducer`


peterhollender:
Interesting - the CI virtual environment isn't picking up the right dependencies (because the "no-option" version grabs a super-minimal set that won't have full functionality). Not sure how to address this? I'd really like `pip install openlifu` to get the primary dependencies with the `[app]` option, but I'm not sure how to make `pyproject.toml` do that

ebrahimebrahim:
> Interesting - the CI virtual environment isn't picking up the right dependencies (because the "no-option" version grabs a super-minimal set that won't have full functionality). Not sure how to address this? I'd really like `pip install openlifu` to get the primary dependencies with the `[app]` option, but I'm not sure how to make `pyproject.toml` do that

I think this is not possible, to make `pip install openlifu` default to `pip install openlifu[app]`. It will always be the minimal thing.

But that is completely fine, in my opinion. It is very reasonable to make the CI ofr example install `pip install openlifu[app]` or whatever we want, and to make the primary instructions be `pip install openlifu[app]` or wahtever. This is common and reasonable.

One way to achieve something like what you are saying is to rename this package to `openlifu-core` and make a new package called `openlifu` that thinly wraps it and that depends on `openlifu-core[app]`. But... eh.

pyproject.toml
```diff
@@ -31,31 +31,54 @@ dynamic = ["version"]
 dependencies = [
   "xarray[io]",
   "numpy<2",
-  "matplotlib",
   "pandas",
   "scipy",
+  "watchdog"
```
ebrahimebrahim: Maybe this one should go to the `cloud` group rather being a core dependency

pyproject.toml
```diff
@@ -31,31 +31,54 @@ dynamic = ["version"]
...
-  "OpenEXR",
-  "scikit-image",
-  "trimesh",
-  "embreex; platform_machine=='x86_64' or platform_machine=='AMD64'",
-  "onnxruntime==1.18.0",
-  "pydicom",
+  "nvidia-ml-py"
+]
+cloud = [
+  "requests",
```
ebrahimebrahim: I think requests should be in the core depndencies group because of its use in misc things like assets.py

It is a very light dependency I think so no issue with that

pyproject.toml
```diff
@@ -86,6 +115,9 @@ Changelog = "https://github.com/OpenwaterHealth/OpenLIFU-python/releases"
 version.source = "vcs"
 build.hooks.vcs.version-file = "src/openlifu/_version.py"
 
+[tool.hatch.metadata]
+allow-direct-references = true
```
ebrahimebrahim: There are no direct references so we should remvoe this. I think for anything merged to main there should not be direct references anyway, because PyPI will reject openlifu if there are

pyproject.toml
ebrahimebrahim, line 14: Problem:
openlifu supports python 3.10+
However openlifu-sdk was set to support 3.12+

This means there will be no dependency version resolution for openlifu on python 3.10 and 3.11!

We need to either bump this up to 3.12 here, giving up on 3.10 and 3.11 forever,
or openlifu-sdk needs to back up and support 3.10, 3.11.

pyproject.toml
```diff
@@ -74,6 +97,12 @@ docs = [
   "sphinx_autodoc_typehints",
   "furo>=2023.08.17",
 ]
```
ebrahimebrahim: The `test` group now needs to be expanded to include the groups that are covered by test. Because the CI installs `.[test]` and so right now it wouldn't have what it needs to run the tests

ebrahimebrahim (changes requested):

dblizinski (approved):

pyproject.toml
peterhollender: Ah, thanks - this was temporarily added to pull an experimental version of the sdk locally, which has since been merged.

pyproject.toml
peterhollender: Is it a potential problem to require 3.12?

ebrahimebrahim (changes requested):
Everything looks good and we [got good results from a hardware enabled test](https://github.com/OpenwaterHealth/SlicerOpenLIFU/pull/565#issuecomment-4632175341), except one remaining issue: openlifu-sdk doesn't actually support python 3.10!

To see the failure, you can try installing `openlifu[io]` from this branch in a Python 3.10 environment, and then try this:

```python
import openlifu_sdk
```

I get

ImportError: cannot import name 'timeout' from 'asyncio'

It's fine in Python 3.12.

Two options here:

- If we want to ditch Python 3.10 compatbility: Release a new openlifu-sdk version with a corrected Requires-Python. Then update the pinned version here to that, and raise the `requires-python` here in openlifu
- If we want to keep Python 3.10: Release a new openlifu-sdk version that replaces asyncio.timeout with a python 3.10 compatible approach. Then update the pinned version here to that.

peterhollender:
> Everything looks good and we [got good results from a hardware enabled test](https://github.com/OpenwaterHealth/SlicerOpenLIFU/pull/565#issuecomment-4632175341), except one remaining issue: openlifu-sdk doesn't actually support python 3.10!
> 
> To see the failure, you can try installing `openlifu[io]` from this branch in a Python 3.10 environment, and then try this:
> 
> ```python
> import openlifu_sdk
> ```
> 
> I get
> 
> ImportError: cannot import name 'timeout' from 'asyncio'
> 
> It's fine in Python 3.12.
> 
> Two options here:
> 
> * If we want to ditch Python 3.10 compatbility: Release a new openlifu-sdk version with a corrected Requires-Python. Then update the pinned version here to that, and raise the `requires-python` here in openlifu
> * If we want to keep Python 3.10: Release a new openlifu-sdk version that replaces asyncio.timeout with a python 3.10 compatible approach. Then update the pinned version here to that.

Hmmm that actually looks like it's not even accessed... I've seen copilot chuck dumb imports in while you are typing before. Can you try removing that line and seeing if it passes? Would be easy to patch to the sdk

ebrahimebrahim:
Removing this line

https://github.com/OpenwaterHealth/openlifu-sdk/blob/7ac69bf618bb98d61f244af06c0f3470804fac69/src/openlifu_sdk/io/LIFUHVController.py#L3

from openlifu-sdk makes openlifu be compatible with python 3.10

Can you guys cut a patch release of openlifu-sdk that removes that line? Then we can update the pinned version here and this should be good to go

ebrahimebrahim (approved):
🥂

PR #450, opened 2026-04-28
Add support for different focal point hit counts OpenwaterHealth/open…

### LINKED ISSUE #449: Add support for non-uniform focal point hit counts
Currently, pulses in a sonication sequence are distributed equally across all focal points in a round robin pattern. In multifocus treatments, focal points can receive different effective acoustic energy per pulse, so equal pulse distribution can leave some foci undertreated relative to ther others. Allowing more pulses to be allocated to weaker foci could raise the min ISPTA across focal points and improve treatment uniformity.

Main proposed changes:

- `Solution.focal_hit_counts`: add a new field storing the number of pulses directed to each focal point
- `Protocol.calc_solution`: add new `focal_hit_counts` and `optimize parameters`. `focal_hit_counts` accepts an explicit allocation. `optimize=True` computes an optimized allocation automatically
- `optimize_hit_counts()`: add a helper that maximizes the min ISPTA across focal points, subject to TIC limits from `param_constraints`
- `SolutionAnalysis.per_focus_tic`: add a new field exposing per-focus TIC values
- weighted intensity aggregation in `calc_solution` and `Solution.get_ispta`

Please let me know if this sounds reasonable and if you'd be open to a PR. I'd love to contribute. Thanks!

samueljwu:
## Summary

Adds support for non-uniform focal point hit counts and `hit_count_optimizer` to maximize minimum ISPTA across foci

Closes #449

## Changes

- Added `Solution.focal_hit_counts` field to store per-focus pulse allocation, validated against number of foci and `sequence.pulse_count`
- Extended `Protocol.calc_solution` with `focal_hit_counts` (explicit allocation) and `optimize` (runs the optimizer) arguments
- Added `optimize_hit_counts()` in `hit_count_optimizer.py`: solves a linear program to maximize the minimum ISPTA across foci, rounds to integers with a local improvement pass, and enforces a hard TIC limit from `param_constraints`
- Added `SolutionAnalysis.per_focus_tic` field
- Weighted TIC and power aggregation in `Solution.analyze()` and per-focus ISPTA weighting in `Solution.get_ispta()` when `focal_hit_counts` is set

## Testing

- Added `tests/test_focal_hit_counts.py` covering:
  - Validation errors for wrong-length or wrong-sum hit counts
  - Correct hit-count-weighted ISPTA and TIC computation
  - Optimizer output sums to `pulse_count` and respects TIC constraints
  - JSON round-trip preserves `focal_hit_counts`
  - Optimizer outperforms random hit count distributions across multiple ISPPA/TIC scenarios
- All existing tests pass
- Pre-commit hooks pass

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -121,37 +121,22 @@ To ensure Meshroom uses your NVIDIA GPU:
 5. Under **Select the preferred graphics processor for this program**, choose **High-performance NVIDIA processor**.
 6. Click **Apply**.
 
-Version control of database using DVC (Data Version Control)
--------------------------------------------------------------
+Getting Sample Data
+-------------------
 
-Data Version Control (DVC) is a data management tool that is meant to be run alongside Git.
-In this project, DVC is used to link changes in the code to specific versions of a sample database containing example project files.
-DVC can be used when this project is installed in Dev mode. You can read more about DVC and how to use it `here <https://dvc.org/doc/start>`_.
-**Note:** Remote access to the sample database stored on google drive is currently restricted. Access requires a :code:`gdrive_client_secret`
-for user access authentication to be shared by developers.
+A sample database for testing and examples is maintained in the
+`openlifu-sample-database <https://github.com/OpenwaterHealth/openlifu-sample-database>`_
+repository. Its files are tracked with Git LFS, so first `install Git LFS
+<https://docs.github.com/en/repositories/working-with-files/managing-large-files/installing-git-large-file-storage>`_.
 
-DVC usage
-~~~~~~~~~
+Then check out the tagged version of the sample database that is compatible
+with this version of ``openlifu``:
 
-To download the sample database:
+.. code:: bash
 
-.. code:: sh
-
-   git pull
-   dvc remote modify --local shared_gdrive gdrive_client_secret <client_secret_here> # Contact developers for grive_client_secret
-   dvc pull # Requires access to remote storage
-
-This will download a directory 'db_dvc' in the repo directory that
-contains the corresponding version of example database files.
-
-To commit updates to the sample database:
-
-.. code:: sh
-
-   dvc add db_dvc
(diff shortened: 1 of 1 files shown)


peterhollender:
Hey @samueljwu this is very cool! I think that the core functionality is quite useful, although I have a suggestion on implementation. I'd like to be conservative about modifying the top level `Solution` attributes, so I'm wondering if this focal-ordering can be added as an optional attribute of `Sequence` instead. 

I'd also like to make the ordering more generic- you are correct to point out that currently, multiple foci default to an `ABCABC...` order. Your proposed ordering makes `AAAABBCCC` possible, but I think we can take this and make it more flexible. If `solution.sequence.focus_order` (I'm open to other names) is a list of indices, then we can represent `AAAABBCCC` as `[0,0,0,0,1,1,2,2,2]`, but could also represent other patterns of arbitrary ordering (`[0,1,2,1,0,1,2,1]` and such). It's not quite as compact for your case (I suppose we could add something like `focus_repeats`, but I'm not overly concerned with the length of the representation in `Solution` since it will be computed). 

Now to implement the computing of the ispta balancing, I'd like to see this integrated into the `solution.scale` method rather than as a separate `optimize` pipeline in 'calc_solution`. Perhaps we introduce a `scaling_options` dict attribute to `Protocol` that will get passed through as `solution.scale(..., **protocol.scaling_options)`. This would also lay the groundwork for other types of balancing (like balancing ispta via amplitude, or doing similar repetition balancing, but with minimized repeats at each position (`ACABCACAB` instead of `AAAABBCCC). 

Sound good?

Great work.

samueljwu:
Thanks @peterhollender for the detailed feedback! This is my first contribution attempt so I really appreciate you taking the time.  This approach makes sense to me and I agree it's cleaner. Let me work through the changes and follow up with an updated version. Excited to contribute here.

PR #451, opened 2026-04-28
Add sample data instructions

ebrahimebrahim:
Replacing DVC by the [openlifu-sample-database](https://github.com/OpenwaterHealth/openlifu-sample-database) as a source of sample data to direct new users to.

I will push tags to openlifu-sample-database that we can reference in the git clone command in the README here. Right now there is just one tag [openlifu-v0.20.0](https://github.com/OpenwaterHealth/openlifu-sample-database/tree/openlifu-v0.20.0) indicating which version of the sample data should be used with openlifu v0.20.0.

Once this is merged, I will updated our release notes for our latest releases to point to this new section in the readme, rather than pointing to dvc data.

For review: Just look at the updated readme and see that the formatting and instructions are sensible.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -121,37 +121,22 @@ To ensure Meshroom uses your NVIDIA GPU:
 5. Under **Select the preferred graphics processor for this program**, choose **High-performance NVIDIA processor**.
 6. Click **Apply**.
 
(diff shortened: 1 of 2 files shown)


sadhana-r (approved):
Looks good to me!

PR #453, opened 2026-05-01
Add CLAUDE.md

ebrahimebrahim:
This is the CLAUDE.md that I've been using for a while, and I now think it makes sense to commit it since we have been using Copilot to help with reviews and apparently [copilot will use CLAUDE.md if it exists](https://docs.github.com/en/copilot/how-tos/copilot-on-github/customize-copilot/add-custom-instructions/add-repository-instructions#creating-custom-instructions). Open to anyone else's CLAUDE.md or AGENTS.md suggestions as well.

If there's any general instructions we want openlifu-python's copilot to always have, now there's a place to put them.


peterhollender (approved):
Great idea

PR #454, opened 2026-05-13
Replace unit handling with Pint OpenwaterHealth/openlifu-python#153

### LINKED ISSUE #153: Better library to handle units
Currently, all physics units are handled in [units.py](https://github.com/OpenwaterHealth/OpenLIFU-python/blob/main/src/openlifu/util/units.py) with a lot of associated function to convert from one unit to another.

Instead we could have it handled by an external library like [Pint](https://pint.readthedocs.io/en/stable/), to profit for upstream work and reduce maintenance burden.

samueljwu:
## Summary
Replaces the unit handling logic in `openlifu.util.units` with Pint implementation while preserving the legacy API and expected behavior for supported units.

Fixes #153 

## Changes
- Added `pint` as a project dependency
- Udpated `units.py` to use `pint.UnitRegistry` for deminsionality checks and conversion factors
- Preserved existing functions
- Improved support for units

All existing tests pass
Pre-commit hooks pass

### DIFF
--- a/src/openlifu/util/units.py
+++ b/src/openlifu/util/units.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import re
+from functools import lru_cache
 
 import numpy as np
 from pint import UnitRegistry
@@ -100,8 +101,35 @@ def _normalize_unit_symbol(unit: str) -> str:
     return unit
 
 
-def _quantity(unit: str):
-    return Q_(1, _normalize_unit(unit))
+def _is_multiplicative(unit: str) -> bool:
+    """Return whether a unit preserves zero when converted to base units."""
+    zero_in_base_units = Q_(0, unit).to_base_units().magnitude
+    return bool(zero_in_base_units == 0)
+
+
+@lru_cache(maxsize=256)
+def _get_conversion_factor(from_unit: str, to_unit: str) -> float:
+    """Return a cached multiplicative conversion factor for a unit pair."""
+    normalized_from_unit = _normalize_unit(from_unit)
+    normalized_to_unit = _normalize_unit(to_unit)
+
+    from_is_angle = normalized_from_unit.lower() in _ANGLE_UNITS
+    to_is_angle = normalized_to_unit.lower() in _ANGLE_UNITS
+
+    # Preserve the legacy distinction between angles and dimensionless values.
+    if from_is_angle != to_is_angle:
+        type0 = getunittype(from_unit)
+        type1 = getunittype(to_unit)
+        raise ValueError(f"Unit type mismatch ({type0}) vs ({type1})")
+
+    converted = Q_(1, normalized_from_unit).to(normalized_to_unit)
+
+    if not (_is_multiplicative(normalized_from_unit) and _is_multiplicative(normalized_to_unit)):
+        raise ValueError(
+            f"Cannot express conversion from {from_unit} to {to_unit} as a multiplicative scale"
+        )
+
+    return float(converted.magnitude)
 
 
 def getunittype(unit):
@@ -134,6 +162,12 @@ def getunittype(unit):
 
 
 def getunitconversion(from_unit, to_unit, unitratio=None, constant=None):
+    """Return the multiplicative scale factor between two units.
+
+    Raises:
+        ValueError: If the units are incompatible, undefined, or cannot be
+            converted using a multiplicative scale factor.
+    """
     if not from_unit:
         return 1.0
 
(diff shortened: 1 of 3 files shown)


ebrahimebrahim:
This is wonderful and much needed, thank you!

Adding @peterhollender to review. I can take it as well, @peterhollender let me know

src/openlifu/util/units.py
```diff
@@ -1,47 +1,147 @@
...
+        return "frequency"
+    if dim == ureg.pascal.dimensionality:
+        return "pressure"
+    if dim == ureg.watt.dimensionality:
+        return "watt"
+
+    return "other"
+
 
 def getunitconversion(from_unit, to_unit, unitratio=None, constant=None):
```
ebrahimebrahim, line 164: There is a limitation of how `openlifu.util.units` works which causes a problem with blind application of pint into it: `getunitconversion` ultimately returns a _multiplier_. This works for units that are on a multiplicative scale,  but consider what happens for units that aren't:

- what would `getunitconversion("degC", "degF")` be? F and C are on an affine scale
- what would `getunitconversion("dB", "dimensionless")` be? dB is on a log scale

We could for now just raise a ValueError for units that are not on a multiplicative scale, to avoid the danger at least

src/openlifu/util/units.py
```diff
@@ -54,129 +154,30 @@ def getunitconversion(from_unit, to_unit, unitratio=None, constant=None):
...
-
-        if slash0 != -1 and slash1 != -1:
-            num0 = from_unit[:slash0]
-            denom0 = from_unit[slash0+1:]
-            num1 = to_unit[:slash1]
-            denom1 = to_unit[slash1+1:]
-            scl = getunitconversion(num0, num1) / getunitconversion(denom0, denom1)
-        elif slash0 == -1 and slash1 == -1:
+        try:
+            scl = _quantity(from_unit).to(_normalize_unit(to_unit)).magnitude
```
ebrahimebrahim: Here we parse from_unit and to_unit repeatedly at every call

I (meaning an AI review agent) ran benchmarks and found that this conversions are 44x slower than the previous implementation.

Since this unit conversion function is called repeatedly per transducer element for some things, we are starting to approach slowdown on the order of 1 second -- starting to get noticeable

Maybe we could cache the conversions? 

so put the `_quantity(from_unit).to(_normalize_unit(to_unit)).magnitude` computation into a helper function and give that caching (e.g. via `functools.cache`) cached by the (from_unit, to_unit) pair

ebrahimebrahim (changes requested):
Thanks for this work, it looks great! Just a couple of changes requested before merge.

(Also I will rebase this branch for you as there has been a lot of change on main since it was put up. So heads up, remember to reset your local)

src/openlifu/util/units.py
```diff
@@ -54,129 +154,30 @@ def getunitconversion(from_unit, to_unit, unitratio=None, constant=None):
...
+        except DimensionalityError as exc:
             type0 = getunittype(from_unit)
             type1 = getunittype(to_unit)
```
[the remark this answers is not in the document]
ebrahimebrahim: copilot does have a point here and feel free to come up with a way to address it, but I think it is fine -- no one is currently relying on angle mismatch raising errors or whether or not angles are dimensionless

samueljwu:
Thanks for checking this out! Summary of the changes:

- Added `_is_multiplicative()` to reject conversions that cannot be represented by a multiplier
- Added `_get_conversion_factor()` helper to avoid repeated conversion
- Added a ValueError for angle to non angle conversions

src/openlifu/util/units.py
```diff
@@ -1,47 +1,181 @@
...
+        prefix, power = watt_match.groups()
+        return f"{prefix}W{power or ''}"
+
+    return unit
+
+
+def _is_multiplicative(unit: str) -> bool:
+    """Return whether a unit preserves zero when converted to base units."""
+    zero_in_base_units = Q_(0, unit).to_base_units().magnitude
+    return bool(zero_in_base_units == 0)
```
ebrahimebrahim, line 107: This is not a mathematically sufficient check which kind of bugs me, but I think you chose a reasonable solution from what was available.

It is acceptable because in Pint (checked 0.24.4 and 0.25.3) there are no unit conversions that preserve 0 but aren't multiplicative. So this check turns out to be sufficient.

The source of truth is pint's own _is_multiplicative metadata whcih suggests the implementation

```py
def _is_multiplicative(unit: str) -> bool:
    return bool(Q_(1, unit)._is_multiplicative)
```

but that's reaching in and messing with a _private_ attribute `_is_multiplicative`.

So I think your approach is acceptable here.

ebrahimebrahim (approved):
This looks great, thank you!

src/openlifu/util/units.py
samueljwu: Thanks for the feedback - I agree. In principle, this check would incorrectly classify a conversion such as y = x^2 as multiplicative. This seems acceptable for now but worth revisiting if there's a better approach available.

PR #471, opened 2026-06-22
Update README.md

stahir715:
Highlighted clicking ok on every tab when adding meshroom to PATH


ebrahimebrahim:
(Add ref to #338 to commit message, or add me as reviewer when ready and I will do it)

PR #472, opened 2026-06-23
Copy pipeline graph to temp dir to avoid whitespaces in path

### LINKED ISSUE #473: Install path with spaces causes problems
It seems to be meshroom related.

Either reproduce and fix the issue, or find a better way to warn users during installation that an path with spaces will cause problems.

sadhana-r:
Meshroom 2025.1.0 has a bug (alicevision/Meshroom#2808) where paths containing spaces are not properly quoted when `meshroom_compute` is launched internally. This causes `meshroom_batch` to fail with unrecognized arguments when the pipeline .mg file lives in a directory with spaces (e.g., OpenLIFU installation dir).

Work around this by copying the pipeline file to the temp directory before passing it to meshroom_batch.

However, this is not a full guarantee if the temp directory itself contains spaces (i.e. if the username has spaces) - although this is usually not the case.

### DIFF
--- a/src/openlifu/nav/photoscan.py
+++ b/src/openlifu/nav/photoscan.py
@@ -422,10 +422,14 @@ def run_reconstruction(
     cache_dir = temp_dir / "MeshroomCache"
     pair_file_path = cache_dir / "imageMatches.txt"
 
+    # Copy pipeline to temp directory without spaces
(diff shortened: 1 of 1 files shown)


ebrahimebrahim (approved):
This looks great, thanks for the investigation to uncover what was going on!

PR #474, opened 2026-07-08
Document release process

ebrahimebrahim:
This is not changing how we do releases; it is just documenting what we already do.

It makes sense to document the process now because we are introducing https://github.com/OpenwaterHealth/SlicerOpenLIFU/pull/581 and https://github.com/OpenwaterHealth/openlifu-desktop-application/pull/87


peterhollender (approved):
Looks good and consistent with our plan

PR #486, opened 2026-07-20
Update release process to use release branches

ebrahimebrahim:
At first I wanted to keep the release process simpler for openlifu-python, but I think that was a mistake on my part.

This updates the release process with one small but important change: we _should_ use release branches. Otherwise the process is the same. Now this brings openlifu-python into parity with the extension and desktop app, which both use release branches.

The amount of textual changes is because I also simplified the document text. Process-wise all that's changing is: _use release branches!_

The reason for the change is: Sometimes when we need to patch the extension of the desktop application the fix needs to take place in openlifu-python. To do that properly we need to use the same same release line of openlifu-python with just a patch fix on top. To do that, we need release branches, and we must never increment the patch version of openlifu-python directly on the main line. 

Related:
- https://github.com/OpenwaterHealth/SlicerOpenLIFU/issues/579
- https://github.com/OpenwaterHealth/openlifu-python/pull/474

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,70 +1,62 @@
 # openlifu-python Release Process
 
-This document describes how to release the `openlifu` Python package. This
-repository is the upstream library for SlicerOpenLIFU and openlifu-app.
+This document describes how to release the `openlifu` Python package. Publishing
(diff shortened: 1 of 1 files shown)


peterhollender (approved):
Makes sense!

PR #488, opened 2026-07-21
Move focal gain LUT to packaged JSON resource

### LINKED ISSUE #448: Save Focal Gain LUT into a data file (json, csv, etc.) to make the source code file shorter.
Totally optional but consider saving the lookup table into a data file (json, csv, etc.) to make the source code file shorter.

If you chose to do that:
- you would read the file here using `importlib.resources` to locate it
- you would ensure it's bundled int he install by adding it to the `[tool.setuptools.package-data]` section. in case hatchling doesn't bnundle it automatically

_Originally posted by @ebrahimebrahim in https://github.com/OpenwaterHealth/openlifu-python/pull/443#discussion_r3066545337_

HussainAther:
Closes #448

## Summary
- move the focal gain lookup table out of `src/openlifu/xdc/util.py`
- store it as a packaged JSON resource
- load it with `importlib.resources`
- preserve the existing `xarray.DataArray` behavior
- include the JSON file in package data

## Testing
- 324 passed
- 2 skipped on macOS as expected


ebrahimebrahim:
This looks good on first glance; planning to do a proper review and integrate this in a couple of weeks if no one gets to it before me. Thank you!

PR #497, opened 2026-08-08
Fix stale volume file on overwrite

### LINKED ISSUE #136: Overwriting a volume can leave a dangling volume file in the database
When using `Databse.write_volume` if we choose to overwrite an existing volume but use a volume file with a different filename, then the old volume file is left there dangling.

HussainAther:
Fixes #136.

When `Database.write_volume()` overwrites an existing volume using a different
source filename, the previously associated volume data file remains in the
database directory.

This change records the existing volume data file before overwrite and removes
it after the replacement file has been copied successfully.

Validation:
- `pytest tests/test_database.py -k issue_136 -v`
- 1 passed, 56 deselected

### DIFF
--- a/src/openlifu/db/database.py
+++ b/src/openlifu/db/database.py
@@ -414,11 +414,17 @@ class Database:
 
         try:
             volume_ids = self.get_volume_ids(subject_id)
+            old_volume_data_filepath = None
             if volume_id in volume_ids:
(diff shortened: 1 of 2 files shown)


ebrahimebrahim:
This looks good on first glance; planning to do a proper review and integrate this in a couple of weeks if no one gets to it before me. Thank you!

